PR #23, opened 2018-04-18
Fixed Vocab location

blester125:
This PR fixes the vocab location. Without this fix the vocab files were not being read so the vocab for each embeddings layer had size 22,270. Now the vocab files are used and embeddings layers are the correct sizes.

### DIFF
--- a/python/mead/config/datasets.json
+++ b/python/mead/config/datasets.json
@@ -48,7 +48,7 @@
 	"label": "atis"
     },
     {
-	"vocab_file": "/data/datasets/iwslt15-en-vi/vocab.en_vi",
+	"vocab_file": "/data/datasets/iwslt15-en-vi/vocab",
 	"train_file": "/data/datasets/iwslt15-en-vi/train",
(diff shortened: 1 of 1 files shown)


blester125:
I might be missing something but doesn't this concatenation solution result in the src and dst vocabs being the same? Is there are reason we want dst words in the src embedding layer? it seems like a waste of memory

dpressel:
No it doesnt.  It uses both and shares the vocabulary.  You may want them in the same vocabulary, especially if you want to tie the weights

dpressel:
If you dont wish to share the vocab, the best option currently is to not provide this file, in which case the `mead` trainer will generate a vocab from source and one from destination separately.

dpressel:
This is a fairly common preprocessing step to merge the vocabularies.  Some refs here:

- https://github.com/EdinburghNLP/nematus/blob/master/data/preprocess.sh
- https://github.com/google/seq2seq/blob/master/bin/data/wmt16_en_de.sh
- https://arxiv.org/pdf/1608.05859.pdf
- https://github.com/EdinburghNLP/nematus

PR #25, opened 2018-04-18
classify fix

blester125:
use reshape over squeeze to that shape info is known


dpressel:
This fixes issue introduced by PR #19

PR #41, opened 2018-04-27
Autodload

sagnik:
- Adds auto download functionality for dataset/ embedding files.
- Three types of downloads are supported:
  - direct link to the file
  - direct link to the file compressed (gz, zip)
  - link to a zipped directory.
-  Docs updated.


python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+
+def mime_type(loc):
+    import magic
+    return magic.Magic(mime=True).from_file(loc)
+
+
+def get_cache_dir():
+    import json
+    try:
+        data_cache = json.load(open("config.json")).get("datacache", "~/.bl-dataset-embeddings/")
```
dpressel: ~/.bl-data?

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        raise RuntimeError("The file can not be downloaded")
+    path_to_save = "/tmp/data.dload-{}".format(os.getpid())
+    try:
+        print("downloading {}".format(url))
+        total_length = None
+        with open(path_to_save, 'wb') as f:
+            total_length = r.headers.get('content-length', None)
+            if total_length is not None:
+                total_length = int(total_length)
+                chunk_size = 1024
```
dpressel: Seems like this chunk size is fairly small.  I would expect that larger chunks would be significantly faster

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+    if os.path.exists(file_name):
+        shutil.rmtree(file_name)
+    return file_name
+
+
+def extract_gzip(file_loc):
+    import gzip
+    import shutil
+    temp_file = get_file("{}.1".format(file_loc))
+    with gzip.open(file_loc, 'rb') as f_in:
```
blester125: You can nest withs like this
```python
with gzip.open(file_loc, 'rb') as f_in, open(temp_file, 'wb') as f_out:
    shutil.copyfileobj(f_in, f_out)
```

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})' # ...or ip
+        r'(?::\d+)?' # optional port
+        r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+    return re.match(regex, url) is not None
+
+
+def update_cache(key):
+    import json
+    try:
+        dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))[key]
```
blester125: This should probably be a shorter name, also saved in a global or something like `CACHE = datasets-embeddings-cache.json`

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        delete_file(file_loc)
+        if key is not None: # cache file validation
+            update_cache(key)
+        return False
+
+
+def check_sanity_dir(dir_loc, dataset_desc, ignore_file_check=False, key=None):
+    import os
+    if not os.path.exists(dir_loc) or not os.path.isdir(dir_loc):
+        update_cache(dir_loc, key)
```
blester125: Above this function only takes a key

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        r'^(?:http|ftp)s?://' # http:// or https://
+        r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|' #domain...
+        r'localhost|' #localhost...
+        r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})' # ...or ip
+        r'(?::\d+)?' # optional port
+        r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+    return re.match(regex, url) is not None
+
+
+def update_cache(key):
```
blester125: How does this function have access to `self`?

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        self.data_download_cache = data_download_cache
+
+    def download(self):
+        import json
+        file_loc = self.dataset_file
+        if check_sanity_file(file_loc):
+            return file_loc
+        elif validate_url(file_loc):  # is it a web URL? check if exists in cache
+            url = file_loc
+            try:
```
blester125: Checking the cache seems like it should be abstracted into a function

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
+import sys
```
blester125: Most python people put all the imports up here rather than in the functions that use them

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+    import magic
+    return magic.Magic(mime=True).from_file(loc)
+
+
+def get_cache_dir():
+    import json
+    try:
+        data_cache = json.load(open("config.json")).get("datacache", "~/.bl-dataset-embeddings/")
+    except IOError:
+        data_cache = "~/.bl-dataset-embeddings/"
```
mtbarta: why not check for the file before loading it instead of silently recovering from an error?

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+                    download_dir = extractor(filepath=temp_file, cache_dir=cache_dir, extractor_func=zipd.get(mime_type(temp_file), None))
+                    if "sha1" in self.dataset_desc:
+                        if os.path.split(download_dir)[-1] != self.dataset_desc["sha1"]:
+                            raise RuntimeError("The sha1 of the downloaded file does not match with the provided one")
+                    dcaches.update({url: download_dir})
+                    json.dump(dcaches, open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json"), "w"), indent=True)
+                    return {k: os.path.join(download_dir, self.dataset_desc[k]) for k in self.dataset_desc if k.endswith("_file")}
+        else:  # we have download links to every file or they exist
+            if not self.enc_dec:
+                return {k: SingleFileDownloader(self.dataset_desc[k], self.data_download_cache).download() for k in self.dataset_desc if k.endswith("_file")}
```
mtbarta: what is the expected behavior of DataDownloader when SingleFileDownloader fails?

SingleFileDownloader throws an exception, which is not handled.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        if check_sanity_file(self.embedding_file):
+            print("embedding file location: {}".format(self.embedding_file))
+            return self.embedding_file
+        try:
+            dacache = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))
+        except IOError:
+            dacache = {}
+        if self.embedding_file in dacache and not self.cache_ignore:
+            download_loc = dacache[self.embedding_file]
+            print("files for {} found in cache".format(self.embedding_file))
```
mtbarta: https://docs.python.org/3/library/logging.html

python/mead/tasks.py
```diff
@@ -5,14 +5,21 @@
...
     TASK_REGISTRY = {}
 
-    def __init__(self, logger_file):
+    def __init__(self, logger_file, mead_config):
         super(Task, self).__init__()
         self.config_params = None
         self.ExporterType = None
+        self.mead_config = mead_config
+        if mead_config is not None:
+            self.data_download_cache = json.load(open(mead_config)).get("datacache", "~/.bl-data-embeddings/")
```
mtbarta: this default is used all over the place. it needs to be defined once.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+class EmbeddingDownloader(Downloader):
+    def __init__(self, embedding_file, embedding_dsz, data_download_cache, cache_ignore=False):
+        super(EmbeddingDownloader, self).__init__(data_download_cache, cache_ignore)
+        self.embedding_file = embedding_file
+        self.embedding_key = embedding_dsz
+        self.data_download_cache = data_download_cache
+
+    def _get_embedding_file(self, loc, key, dload_url):
```
mtbarta: what is this supposed to return? L 259 returns a file location, L 261 returns a new copy of this class.

mtbarta: if this method should return an instance of EmbeddingDownloader, then it should be marked as a class method.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+                print("file for {} found in cache, not downloading".format(url))
+                return dcaches[url]
+            else:  # download the file in the cache, update the json
+                cache_dir = self.data_download_cache
+                print("using {} as data/embeddings cache".format(cache_dir))
+                zipd = {'application/gzip': extract_gzip, 'application/zip': extract_zip}
+                temp_file = web_downloader(url)
+                dload_file = extractor(filepath=temp_file, cache_dir=cache_dir, extractor_func=zipd.get(mime_type(temp_file), None))
+                dcaches.update({url: dload_file})
+                json.dump(dcaches, open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json"), "w"), indent=True)
```
mtbarta: file name shouldn't be hardcoded here.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+            try:
+                dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))
+            except IOError:
+                dcaches = {}
+            if dload_bundle in dcaches and check_sanity_dir(dcaches[dload_bundle], self.dataset_desc, self.enc_dec, dload_bundle) and not self.cache_ignore:
+                download_dir = dcaches[dload_bundle]
+                print("files for {} found in cache, not downloading".format(dload_bundle))
+                return {k: os.path.join(download_dir, self.dataset_desc[k]) for k in self.dataset_desc if k.endswith("_file")}
+            else:  # try to download the bundle and unzip
+                url = dload_bundle
```
mtbarta: this feels like an unnecessary indirection. If url is more descriptive of the value, then change the name of dload_bundle.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+                dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))
+            except IOError:
+                dcaches = {}
+            if url in dcaches and check_sanity_file(dcaches[url], url) and not self.cache_ignore:
+                print("file for {} found in cache, not downloading".format(url))
+                return dcaches[url]
+            else:  # download the file in the cache, update the json
+                cache_dir = self.data_download_cache
+                print("using {} as data/embeddings cache".format(cache_dir))
+                zipd = {'application/gzip': extract_gzip, 'application/zip': extract_zip}
```
mtbarta: pull this LUT outside. it only needs to be defined once.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
+import sys
+
+def get_file(file_name):
```
mtbarta: method repeats delete_file().

this method should probably not be deleting anything, given the name.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        r'(?::\d+)?' # optional port
+        r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+    return re.match(regex, url) is not None
+
+
+def update_cache(key):
+    import json
+    try:
+        dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))[key]
+    except (IOError, KeyError):
```
mtbarta: i think we can avoid throwing errors here by refactoring L122.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+    if os.path.exists(file_loc):
+        shutil.rmtree(file_loc)
+
+
+def check_sanity_file(file_loc, key=None):
+    import os
+    if os.path.exists(file_loc) and os.path.isfile(file_loc) and not mime_type(file_loc) == "text/html":  # dropbox doesn't give 404 in case the file does not exist, produces an HTML. Almost never the actual train/dev/test file is an HTML.
+        return True
+    else:
+        delete_file(file_loc)
```
mtbarta: os.path.exists() is called twice when the file doesn't exist.

it's surprising to me that a method called `check_sanity_file` would delete something.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def download(self):
+        import json
+        file_loc = self.dataset_file
+        if check_sanity_file(file_loc):
+            return file_loc
+        elif validate_url(file_loc):  # is it a web URL? check if exists in cache
+            url = file_loc
+            try:
+                dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))
```
mtbarta: fairly certain opening files requires a call to close() to free up system resources.

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+        self.data_download_cache = data_download_cache
+        self.enc_dec = enc_dec
+
+    def download(self):
+        import os
+        import json
+        dload_bundle = self.dataset_desc.get("download", None)
+        if dload_bundle is not None:  # download a zip/tar/tar.gz directory, look for train, dev test files inside that.
+            try:
+                dcaches = json.load(open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json")))
```
mtbarta: does this file not need to be closed? see above

python/mead/downloader.py
```diff
@@ -0,0 +1,300 @@
...
+                else:
+                    cache_dir = self.data_download_cache
+                    temp_file = web_downloader(url)
+                    zipd = {'application/gzip': extract_gzip, 'application/zip': extract_zip}
+                    download_dir = extractor(filepath=temp_file, cache_dir=cache_dir, extractor_func=zipd.get(mime_type(temp_file), None))
+                    if "sha1" in self.dataset_desc:
+                        if os.path.split(download_dir)[-1] != self.dataset_desc["sha1"]:
+                            raise RuntimeError("The sha1 of the downloaded file does not match with the provided one")
+                    dcaches.update({url: download_dir})
+                    json.dump(dcaches, open(os.path.join(self.data_download_cache, "datasets-embeddings-cache.json"), "w"), indent=True)
```
mtbarta: how many times do we open the same file?

python/mead/downloader.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def check_sanity_dir(dir_loc, dataset_desc, data_dcache=None, ignore_file_check=False, key=None):
+    if not os.path.exists(dir_loc) or not os.path.isdir(dir_loc):
+        update_cache(key, data_dcache)
+        return False
+    if ignore_file_check:  # for enc_dec tasks there's no direct downloads
+        return True
+    files = [os.path.join(dir_loc, dataset_desc[k]) for k in dataset_desc if k.endswith("_file")]
+    for file in files:
```
dpressel: use a different var name (f?)

python/mead/downloader.py
```diff
@@ -0,0 +1,287 @@
...
+
+def update_cache(key, data_download_cache):
+    dcache = read_json(os.path.join(data_download_cache, DATA_CACHE_CONF))
+    if key not in dcache:
+        return
+    del dcache[key]
+    write_to_json(dcache, os.path.join(data_download_cache, DATA_CACHE_CONF))
+
+
+def check_sanity_file(file_loc, data_dcache=None, key=None):
```
dpressel: Name of this function could be clearer.

python/mead/downloader.py
```diff
@@ -0,0 +1,287 @@
...
+
+def read_json(filepath):
+    if not os.path.exists(filepath):
+        return {}
+    with open(filepath) as f:
+        j_con = json.load(f)
+    return j_con
+
+
+def write_to_json(content, filepath):
```
dpressel: write_json

python/mead/downloader.py
```diff
@@ -0,0 +1,287 @@
...
+    def download(self):
+        pass
+
+
+class SingleFileDownloader(Downloader):
+    def __init__(self, dataset_file, data_download_cache, cache_ignore=False):
+        super(SingleFileDownloader, self).__init__(data_download_cache, cache_ignore)
+        self.dataset_file = dataset_file
+        self.data_download_cache = data_download_cache
+        self.zipd = {'application/gzip': extract_gzip, 'application/zip': extract_zip}
```
dpressel: Appears to be shared between all impls.  Push up.  Also seems that it could be defined as a static

python/mead/downloader.py
```diff
@@ -0,0 +1,310 @@
...
+        self.dataset_desc = dataset_desc
+        self.data_download_cache = data_download_cache
+        self.enc_dec = enc_dec
+
+    def download(self):
+        dload_bundle = self.dataset_desc.get("download", None)
+        if dload_bundle is not None:  # download a zip/tar/tar.gz directory, look for train, dev test files inside that.
+            dcache_path = os.path.join(self.data_download_cache, DATA_CACHE_CONF)
+            dcache = read_json(dcache_path)
+            if dload_bundle in dcache and \
```
mtbarta: i'm wondering if it would be easier to read this line if `is_dir_correct` was wrapped in an internal method so that the instance variables don't have to be passed to it.

docs/dataset-embedding.md
```diff
@@ -0,0 +1,119 @@
...
+        "label": "glove-6B-50",
+        "file": "https://www.dropbox.com/s/339mhx40t3q9bp5/glove.6B.50d.txt.gz?dl=1",
+        "dsz": 50
+    },
+```
+
+- On file system
+
+```aidl
+{
```
blester125: Formatting on these verbatim sections should be consistant

python/mead/downloader.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+class DataDownloader(Downloader):
+    def __init__(self, dataset_desc, data_download_cache, enc_dec=False, cache_ignore=False):
+        super(DataDownloader, self).__init__(data_download_cache, cache_ignore)
+        self.dataset_desc = dataset_desc
+        self.data_download_cache = data_download_cache
+        self.enc_dec = enc_dec
+
+    def download(self):
```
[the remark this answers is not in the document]
dpressel: Well, last else is not needed since all other paths return

python/mead/downloader.py
```diff
@@ -0,0 +1,310 @@
...
+    with zipfile.ZipFile(file_loc, "r") as zip_ref:
+        zip_ref.extractall(temp_file)
+    return temp_file
+
+
+def mime_type(loc):
+    return magic.Magic(mime=True).from_file(loc)
+
+
+zipd = {'application/gzip': extract_gzip, 'application/zip': extract_zip}
```
dpressel: This shouldnt be a global

python/mead/downloader.py
```diff
@@ -0,0 +1,310 @@
...
+        super(EmbeddingDownloader, self).__init__(data_download_cache, cache_ignore)
+        self.embedding_file = embedding_file
+        self.embedding_key = embedding_dsz
+        self.data_download_cache = data_download_cache
+        self.sha1 = embedding_sha1
+
+    @staticmethod
+    def _get_embedding_file(loc, key):
+        if os.path.isfile(loc):
+                LOG("embedding file location: {}".format(loc))
```
blester125: this is too indented

PR #42, opened 2018-05-02
Set up pip install

blester125:
This PR makes baseline, mead, and xpctl pip installable. 

A test package can be installed with `pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple --no-cache-dir deep-baseline`

I tested it with python 2.7, 3.5, and 3.6 where I created new environments, installed tensorflow or pytorch and pip installed baseline then ran some experiments with `mead-train --config config/sst2.json`. Testing was done with both backends and after merge of the autodownloader so that part works too.

I also exported the tensorflow models for each python version with `mead-export --config config/sst2.json --model classify-model-tf-{PID}`, served the exported model with tfserving, and sent requests to it to verify that it worked.

Current behavior of arguments to mead is that is the file specified is found that is used but if it not then it searches for that path relative to where mead is installed. This lets use use the example configs which come bundled with the release. This can be tweaked if we need to.

Also on pypi there is already a package called baseline which had it's last release today so it doesn't look we can get that name.

This PR also includes some docs updates to that they are easier to link to from PyPi.

Having 2 `setup.py` in the same directory is not a use case they normally cover so there is a driver program to build distributions. You run it like `python release.py (xpctl|baseline)` then you can upload the results to twine

I also tested addon behavior. When a user has a directory in their python path with an addon in it it gets imported and used. When the addon if from our repo it is possible to import that in the config. When there is a name clash (the user has an addon with the same name as ours) the user one gets picked over ours.


xpctl/README.md
```diff
@@ -1 +0,0 @@
-See the [docs](../docs/xpctl.md)
```
sagnik: ll this remove the README from the `xpctl` folder?

blester125: Yeah, the `xpctl` folder is now only code but the `docs/xpctl.md` is linked as the README on pypi

PR #44, opened 2018-05-04
remove magic dependency

mtbarta:


dpressel:
Went with @blester125's mod

PR #60, opened 2018-06-01
default to cwd if model argument is not path-like

mtbarta:
* finding vocab files always returns absolute paths.
* finding vocab suffixes works by splitting the path and then subtracting the file name.

### DIFF
--- a/python/baseline/tf/tfy.py
+++ b/python/baseline/tf/tfy.py
@@ -15,16 +15,30 @@ def _find_files_by_type(model_file, filetype):
     matching_files = []
 
     filetype_ending = "." + filetype
-    basepath = os.path.dirname(model_file)
-    if not os.path.isdir(basepath):
-        raise IOError("not a model directory")
+    basepath = get_basepath_or_cwd(model_file)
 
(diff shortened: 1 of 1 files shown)


python/baseline/tf/tfy.py
```diff
@@ -15,17 +15,31 @@ def _find_files_by_type(model_file, filetype):
...
-            matching_files.append(filename_without_ending)
+            matching_files.append(os.path.join(basepath, filename_without_ending))
+
+    if not matching_files:
+        raise ValueError("no vocab files found in directory %s. \
+Please specify the model as path-like. e.g. /data/model/model-name-1234" % basepath)
 
     return matching_files
 
+def get_basepath_or_cwd(model_file):
```
dpressel: put to utils.py?

mtbarta: if this belongs in utils.py, would _find_files_by_type also belong there?

dpressel: will merge for now, can figure out later if we should pull it up

PR #63, opened 2018-06-04
Added EMA training, loading, and exporting to tensorflow and tests

blester125:
The PR add Exponential Moving Average training to Tensorflow Classify.

The actual ema operations are added with the function `_add_ema` in tfy, ema calculation is done with a `tf.control_dependencies` in the trainer, and using the ema values is done with a special `tf.train.Saver` that reads the EMA values into the variable tensors.

When saving the model a new `.eval_saver` is created. This can be used when restoring the model.

This is how the EMA works for training:
 * When training the model it creates a checkpoint at the end of each epoch. This has the current params and the ema version. 
 * Before testing this checkpoint is restored using the special saver.
 * Evaluation is performed and the model is checkpointed for real if the performance is better.
 * The training checkpoint it then loaded and training continues.

This format allows for reloading the whole model for retraining, reload for patience, and things like that.

This PR also modifies the exporter so that if there is an `.eval_saver` present it loads in the EMA values.

The exporting was tested by exporting both EMA and normal models to make sure they work as well an an automated test that compares the saved values from a model trained with EMA.

The EMA is tested with many manual runs as well as automated testings that covers things like the emas are in the graph, get updated, can be saved and restored and the like.

There was also a small change to the exporter that allows for it to export a model created with random word vectors. This was tested by exporting untrained models with random vectors and pretrained and checking that the values in the embedding table were correct


blester125:
Solved a different way with #109

PR #65, opened 2018-06-08
Variational Dropout for Pytorch

blester125:
Adds a Variational Dropout (same mask at each time-step) to Pytorch.

This only works on the input/output of the RNN, not on recurrent connections yet. 

Assumes that input is `(T, B, H)` which is what RNNs already do to to utilize cudnn.

Tagger using Variational Dropout is currently top of tagger conll results in xpctl

### DIFF
--- a/python/baseline/tf/classify/model.py
+++ b/python/baseline/tf/classify/model.py
@@ -503,6 +503,7 @@ class LSTMModel(WordClassifierBase):
         :return: 
         """
         hsz = kwargs.get('rnnsz', kwargs.get('hsz', 100))
+        vdrop = bool(kwargs.get('variational_dropout', False))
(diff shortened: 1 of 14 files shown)


python/baseline/tf/classify/model.py
```diff
@@ -389,9 +389,10 @@ def pool(self, word_embeddings, dsz, init, **kwargs):
         :return: 
         """
         hsz = kwargs.get('rnnsz', kwargs.get('hsz', 100))
+        vdrop = kwargs.get('varitational_dropout', False)
```
dpressel: typo

blester125:
I trained a model on master then checkout this branch and I was able to load and run the model

PR #73, opened 2018-06-14
Calculate norms in loop and compare to vectorized for testing.

blester125:
This PR does the following.

 * Make norm testing more explicit with per vector norm in loop for calculating gold. 
 * It filters out repeated words in the embeddings files
 * Fixes non breaking spaces getting skipped in Glove and Word2Vec


dpressel:
I tested this change against SST2 and CONLL2003.  No regressions seen in performance.  Will merge this soon

PR #79, opened 2018-06-19
optionally save confusion matrix to csv

sagnik:


python/baseline/confusion.py
```diff
@@ -48,6 +50,16 @@ def __str__(self):
         values += ['\n']
         return ''.join(values)
 
+    def write_to_file(self, outfile):
```
dpressel: I think `save()` would be more succinct

python/baseline/dy/classify/train.py
```diff
@@ -48,12 +48,11 @@ def _step(self, loader, update, verbose=False):
 
         metrics = cm.get_all_metrics()
         metrics['avg_loss'] = total_loss / float(steps)
-        if verbose:
-            print(cm)
+        verbose_output_classify(verbose, cm)
```
dpressel: This function has a kind of imprecise name, as it implies that some sort of classification is performed when, in fact, its more of a reporting/output behavior.

PR #80, opened 2018-06-22
Created CRF Valid transition masks and let you apply them to models

blester125:
This PR adds functions to create legal CRF transition masks for `IOB` `IOB2` and `IOBES`.

Masks are applied to the CRF's when the `crf_mask` parameter is supplied to the `model` section of the model config and a `span_type` parameter is supplied to the `train` section. This is not the best interface and I think we should discuss how to set up the parameters so that it is easy to use/not use for end users.

Masks are created once and then applied to the weights before the CRF runs rather than just initializing with the mask.

This mask work by adding very negative values to the transition parameters themselves rather than masking the scores with zero before the logsumexp. Currently it is done this way because it is consistent between train and test and because we are using the contrib crf in tesnorflow it is the only way to mask it for tf.

Masks are column based (`x->y` is `T[y, x]`) because that is how Pytorch and Dynet CRFs work. Tensorflow is row based so we transpose the mask.

This also includes tests for the mask and makes sure that it is applied in the full models.

In Tensorflow the CRF contrib module doesn't use a `<GO>` state for the initial state so the mask wasn't stopping the model from emitting `I-` or `E-` as the first output. A new fake state is not pre-pended to the input so the mask does work. I tested this and it seems not to break backward compatibility. I was able to take a model trained from master and reload/export it with the new code and the model still worked.


python/baseline/pytorch/tagger/model.py
```diff
@@ -80,7 +83,14 @@ def create(labels, embeddings, **kwargs):
             ))
 
         if model.use_crf:
-            model.crf = CRF(len(labels), (model.labels.get("<GO>"), model.labels.get("<EOS>")))
+            if model.span_type is not None:
+                model.crf = CRF(
+                    len(labels),
+                    (model.labels.get("<GO>"), model.labels.get("<EOS>")),
+                    model.labels, model.span_type, model.labels.get("<PAD<")
```
dpressel: typo

dpressel:
Testing and reviewing.

python/baseline/pytorch/tagger/model.py
```diff
@@ -12,6 +12,8 @@ def save(self, outname):
     def to_gpu(self):
         self.gpu = True
         self.cuda()
+        if self.use_crf:
```
dpressel: Remove?

dpressel: should be cascading from parent, right?

python/mead/tasks.py
```diff
@@ -331,6 +331,7 @@ def initialize(self, embeddings):
 
     def _create_model(self):
         labels = self.reader.label2index
+        self.config_params['model']['span_type'] = self.config_params['train'].get('span_type')
```
dpressel: this wont set `crf_mask` so Im not sure how that property gets set

python/mead/tf/exporters.py
```diff
@@ -306,6 +306,7 @@ def _run(self, sess, model_file, embeddings_set, output_dir, model_version):
         model_params["sess"] = sess
         model_params["maxs"] = mxlen
         model_params["maxw"] = mxwlen
+        model_params['span_type'] = self.task.config_params['train'].get('span_type')
```
dpressel: See previous comment, doesnt set `crf_mask` which is the kwarg that model uses

PR #85, opened 2018-07-05
Batched CRF in pytorch

blester125:
This is a batched version of a CRF in pytorch. When I was testing the CRF masks in pytorch each epoch was way too slow so I batched it.

```
Old Speed:
Train: 112.7518 ± 1.064
Dev:     4.4706 ± 0.023651

Batched Speed:
Train:   37.0712 ± 0.143
Dev:      2.1887 ± 0.009753
```

Tested by running examples through the old and the new CRF and making sure the values match. Also I tested changes to the tagger model by training a tagger with CRF and mask, with CRF and no mask, and no CRF. All models successfully trained.


dpressel:
Testing this

dpressel:
I tested this with 10 runs each of old and new. Verified no regression of performance!

PR #100, opened 2018-08-08
Xpctl tasksummary

sagnik:
- **commands**
  - **syntax change**: `results <taskname> <datasetname>`. Provides a statistical summary of all experiments on that dataset, grouped by config sha1s. can be filtered on user(s), metric(s), number of experiment(s). can be sorted. shows results on the test data by default. provide the event_type option to see results on train/valid.
  - **addition**: `details <taskname> <sha1>`. Provides the detailed results for _one_experiment. can be filtered on user(s), metric(s), number of experiment(s). can be sorted. shows results on the test data by default. provide the event_type option to see results on train/valid. 
  - **deletion**: `best`, `tasksummary`.
- **test coverage**: mongo + postgresql.
- **docs updated**.

### DIFF
--- a/python/mead/config/conll-iobes.json
+++ b/python/mead/config/conll-iobes.json
@@ -30,7 +30,7 @@
         "label": "glove-6B-100"
     },
     "train": {
-        "epochs": 100,
+        "epochs": 2,
         "optim": "sgd",
         "decay": 0.05,
         "eta": 0.015,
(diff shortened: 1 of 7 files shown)


python/mead/config/conll-iobes.json
```diff
@@ -30,7 +30,7 @@
         "label": "glove-6B-100"
     },
     "train": {
-        "epochs": 100,
+        "epochs": 2,
```
dpressel: This shouldnt change!

sagnik: fixed

PR #106, opened 2018-08-15
abstract saved model builder. added tensor signature

mtbarta:
A user should be able to send raw tensors over the network.

In order to achieve this, I pulled out the saved model builder into a common method and added
an endpoint that accepts tensors as input with the standard output definition.

Names are up for debate.


python/tests/test_exporter.py
```diff
@@ -0,0 +1,136 @@
+import pytest
+import numpy as np
+from unittest.mock import patch, Mock
```
blester125: `patch` and `Mock` should be imported from the mock library (the python2.7 backport)

PR #114, opened 2018-08-23
Load and save tensorflow models as zip

sagnik:
- Load and save tensorflow models as zip
- some errors that seemed obvious

### DIFF
--- a/python/baseline/utils.py
+++ b/python/baseline/utils.py
@@ -745,19 +745,17 @@ def f_score(overlap_count, gold_count, guess_count, f=1):
 
 @exporter
 def unzip_model(path):
-    from mead.mime_type import mime_type
     """If the path for a model file is a zip file, unzip it in /tmp and return the unzipped path"""
-    if not os.path.exists(path) or not mime_type(path) == "application/zip":
-        return path
-    with open(path, 'rb') as f:
-        sha1 = hashlib.sha1(f.read()).hexdigest()
-    temp_dir = os.path.join("/tmp/", sha1)
-    if not os.path.exists(temp_dir):
-        print("unzipping model before exporting")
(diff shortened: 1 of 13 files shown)


python/baseline/utils.py
```diff
@@ -741,3 +743,31 @@ def f_score(overlap_count, gold_count, guess_count, f=1):
     f = (1. + beta_sq) * (precision * recall) / (beta_sq * precision + recall)
     return f
 
+@exporter
+def unzip_model(path):
+    from mead.mime_type import mime_type
```
dpressel: Need to pull mime type into bl utils for this change. Don’t want to require a dependency from bl to mead package

python/baseline/utils.py
```diff
@@ -741,3 +743,31 @@ def f_score(overlap_count, gold_count, guess_count, f=1):
...
+    return path
+
+
+@exporter
+def zip_model(path):
+    """zips the model files"""
+    print("zipping model files")
+    model_files = [x for x in os.listdir(".") if path[2:] in x]
+    z = zipfile.ZipFile("{}.zip".format(path), "w")
+    for file in model_files:
```
dpressel: Let’s define var as _f_

python/tag.py
```diff
@@ -33,7 +31,7 @@ def read_lines(tsfile):
                                              'predicted label')
 parser.add_argument('--input', help='input conll', required=True)
 parser.add_argument('--output', help='output conll', required=True)
-parser.add_argument('--model', help='model file', required=True, type=unzip_model)
+parser.add_argument('--model', help='model file', required=True)
```
dpressel: It would seem like other programs for eval also require this change

PR #119, opened 2018-08-24
Updated Visdom docs

blester125:


dpressel:
I think I updated with all pertinent info.  LMK if not the case

PR #130, opened 2018-09-14
Update reporting.md with more context

amyhemmeter:
Giving a little more context for the reporting hook in the config (i.e. showing that it's not part of train).

### DIFF
--- a/docs/reporting.md
+++ b/docs/reporting.md
@@ -4,7 +4,7 @@ A `reporting hook` for baseline is a class that has three methods: `init`, `step
 
 You can use the reporting hooks in two ways: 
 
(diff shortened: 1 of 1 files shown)


docs/reporting.md
```diff
@@ -4,13 +4,24 @@ A `reporting hook` for baseline is a class that has three methods: `init`, `step
 
 You can use the reporting hooks in two ways: 
 
-- **mead config file**: The config file should have a field called "reporting". Eg: 
+- **mead config file**: The config file should have a field called "reporting" at the end of the json config. Eg:
```
sagnik: doesn't have to be at the end.

PR #133, opened 2018-09-20
Dynet Docker and automatic speed tests.

blester125:
This PR adds tools for running and aggregating results from speed tests in baseline. It records information about the runs to group them and orders display based on the recency of the used frameworks. It also generates a markdown file with the information in it.

Future improvements:

 * Allow for a remote database so that multiple people can submit times
 * Find a clean way to order GPUs so that results from beefer GPUs show up first.


scripts/speed_test/run.py
```diff
@@ -0,0 +1,423 @@
...
+
+def version_str_to_tuple(version):
+    m = simple_version_regex.match(version)
+    if m is None:
+        return Version(0, 0, 0)
+    return Version(int(m.group(1)), int(m.group(2)), int(m.group(3)))
+
+def get_cuda_version():
+    """Find the version of CUDA and CUDNN."""
+    cuda_version = open('/usr/local/cuda/version.txt').read().split()[-1]
```
dpressel: check `CUDA_HOME`

PR #134, opened 2018-09-20
Use a real softmax op

blester125:
In the old model a new `tf.nn.softmax` was created for each call to `classify`. This PR created a dedicated softmax op and runs with that instead.

A quick test running through the SST2 dev set calling `classify_text` took 80s with the old way. The new way takes 3s


python/baseline/tf/classify/model.py
```diff
@@ -403,6 +404,7 @@ def create(cls, embeddings, labels, **kwargs):
                 with tf.variable_scope("output"):
                     model.logits = tf.identity(fully_connected(stacked, nc, activation_fn=None), name="logits")
                     model.best = tf.argmax(model.logits, 1, name="best")
+                    model.probs = tf.nn.softmax(model.logits, name="probs")
```
dpressel: For backwards compat, I would think that `load()` needs to be updated to create this if its not found

blester125: Fixed. tested by training a model on master and reloading it in this branch

PR #139, opened 2018-09-25
Move utils

blester125:
This PR moves utilities for hashing configs into baseline. It also adds some unittests and switches xpctl to using these for hashing.

I was able to train models and save results to local xpctl so it doesn't seem broken.

### DIFF
--- a/python/baseline/utils.py
+++ b/python/baseline/utils.py
@@ -2,22 +2,25 @@ import os
 import sys
 import json
 import logging
+import hashlib
+import zipfile
 import importlib
+from copy import deepcopy
+from collections import OrderedDict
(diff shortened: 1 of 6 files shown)


python/baseline/utils.py
```diff
@@ -829,3 +836,84 @@ def verbose_output(verbose, confusion_matrix):
...
+    for (key, value) in sorted(data.items(), key=lambda x: x[0]):
+        if isinstance(value, dict):
+            value = order_json(value)
+        elif isinstance(value, list):
+            value = sorted(value)
+        new[key] = value
+    return new
+
+
+KEYS = {
```
dpressel: This seems like it belongs in mead, not baseline

PR #144, opened 2018-10-03
exporting in v1. read in features from and vocab objects up front.

mtbarta:
A user should be able to export from v1 mead.

mead/export.py reads in feature objects before running the exporter. We reassemble the correct vocab and weight objects in the exporter.

baseline's tf/embeddings.py will default the weights of character embeddings.

To support instantiating different embedding classes, I use a star import on tf embeddings. There may be a better way, but I also wanted to avoid an `eval()` inside the exporter.

### DIFF
--- a/python/mead/tf/exporters.py
+++ b/python/mead/tf/exporters.py
@@ -6,11 +6,13 @@ import os
 from tensorflow.python.framework.errors_impl import NotFoundError
 import mead.utils
 import mead.exporters
+from baseline.tf.embeddings import *
 from mead.tf.signatures import SignatureInput, SignatureOutput
 from mead.tf.preprocessor import PreprocessorCreator
 from baseline.utils import export
 from baseline.tf.tfy import get_vocab_file_suffixes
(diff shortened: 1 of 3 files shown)


python/mead/tf/exporters.py
```diff
@@ -186,18 +207,16 @@ def _create_preprocessed_input(self, tf_example,
...
-            mxwlen = state['maxw']
-        if 'mxwlen' in state:
-            mxwlen = state['mxwlen']
-        return mxlen, mxwlen
+    def _get_max_lens(self, feature_descs):
+        """
+        _get_max_lens() restructures the feature_descs object
+        to order them for mxlen and mxwlen, which is word and char.
+        """
+        if 'word' not in feature_descs:
```
dpressel: Its actually possible to not have a feature named 'word' or 'char' at all.  There is a concept of a primary feature when mead runs now, and the vectorizers also track both the mxlen/mxwlen as well as any user-defined length constraints, so I think this can be generalized.

PR #146, opened 2018-10-07
fix reporting error

sagnik:


blester125:
This error happens when you have something like `"reporting": ["xpctl"]` in the mead config?

`reporting = {k: {} for k in reporting} if isinstance(reporting, list) else reporting`

blester125:
#155 fixes this in a different way

PR #150, opened 2018-10-11
Plugin dec

blester125:
This PR adds a plugin decorator that creates a create and load hook for the decorated class.

When the class has a `create` classmethod that is used, otherwise it calls the constructor directly.

When the class has a `load` classmethod that is added.

You can pass a name to it so that instead of using `create_model` you can have `create_reporting_hook` or `create_embeddings`. Do we want to have different hooks for different types of addons? Seems like just a create/load would be simpler.

### DIFF
--- a/python/addons/reporting_xpctl.py
+++ b/python/addons/reporting_xpctl.py
@@ -6,7 +6,7 @@ import socket
 import os
 
 
-@plugin('create_reporting_hook')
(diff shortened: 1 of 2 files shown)


dpressel:
Ok, I think the key parts have been covered in the v1, so this one can be closed out

PR #161, opened 2018-10-22
Feature/remote services

mtbarta:
A user should be able to call a service remotely.

This PR will introduce logic to transform tensors to TensorProtos and send them over gRPC to tensorflow serving within the Services.

* exporters copy vectorizers, vocabs, and labels to export bundle
* exporters save asset file with model variables for client use
* Services can use RemoteModel to connect to remote models
* RemoteModel reads in asset file to find labels, signature inputs, and lengths
* s2s exporter sets beam with to 30
* vectorizers use int32 types


python/baseline/tf/seq2seq/model.py
```diff
@@ -466,7 +466,7 @@ def predict(self, batch_dict):
         if len(vec.shape) == 3:
             return vec.transpose(1, 2, 0)
         else:
-            return vec.transpose(1, 0)
+            return [vec.transpose(1, 0)]
```
dpressel: Lets not change anything from this class. Its getting refactored.   Assume for now a 3D tensor, as that will be the case after my latest mods

api-examples/ed-text.py
```diff
@@ -21,5 +22,5 @@
 
 print(texts)
 
-m = bl.EncoderDecoderService.load(args.model, backend=args.backend)
-print(m.predict(texts))
+m = bl.EncoderDecoderService.load(args.model, backend=args.backend, remote=args.remote, name=args.name)
+print(m.transform(texts))
```
dpressel: `transform` was changed to `predict` -- this undoes that

python/baseline/model.py
```diff
@@ -234,3 +234,106 @@ def predict(self, source_dict, **kwargs):
...
+        self.signature = signature
+
+        self.channel = self.grpc.insecure_channel(remote)
+
+        self.lengths_key = lengths_key
+        self.input_keys = set(inputs)
+        self.beam = beam
+        self.labels = labels
+
+    def classify(self, examples):
```
dpressel: These can probably go away

dpressel: Each model now has a `predict` method

python/baseline/model.py
```diff
@@ -234,3 +234,106 @@ def predict(self, source_dict, **kwargs):
...
+    def predict(self, examples):
+        return self._transform(examples)
+
+    def predict_next(self, examples):
+        return self._transform(examples)
+
+    def get_labels(self):
+        return self.labels
+
+    def _transform(self, examples):
```
dpressel: this -> `predict`

python/baseline/model.py
```diff
@@ -234,3 +234,106 @@ def predict(self, source_dict, **kwargs):
...
+            classes = predict_response.outputs.get('classes').int_val
+            results = [classes[x:x+self.beam] for x in range(0, len(classes), self.beam)]
+            results = list(zip(*results)) #transpose
+            return [results]
+
+        if self.signature == 'tag_text':
+            classes = predict_response.outputs.get('classes').int_val
+            lengths = examples[self.lengths_key]
+            result = []
+            for i in range(examples['word'].shape[0]):
```
dpressel: `word` is not a required feature in tagging.  This wont work generally

python/baseline/services.py
```diff
@@ -254,7 +303,7 @@ def predict(self, tokens, **kwargs):
                     else:
                         examples[lengths_key] = np.array(length)
             batch_dict = {k: v.reshape((1,) + v.shape) for k, v in examples.items()}
-            softmax_tokens = self.model.predict(batch_dict)
+            softmax_tokens = self.model.predict_next(batch_dict)
```
dpressel: swap these

python/baseline/services.py
```diff
@@ -97,9 +145,9 @@ def predict(self, tokens):
 
         for k in self.vectorizers.keys():
             examples[k] = np.stack(examples[k])
-            lengths_key = '{}_lengths'.format(k)
-            examples[lengths_key] = np.stack(examples[lengths_key])
-        outcomes_list = self.model.predict(examples)
+
+        outcomes_list = self.model.classify(examples)
```
dpressel: Call predict

python/baseline/services.py
```diff
@@ -224,7 +280,7 @@ def load(cls, bundle, **kwargs):
         return super(LanguageModelService, cls).load(bundle, **kwargs)
 
     # Do a greedy decode for now, everything else will be super slow
-    def predict(self, tokens, **kwargs):
+    def run(self, tokens, **kwargs):
```
dpressel: This is incorrect, should be `predict`

python/baseline/tf/remote.py
```diff
@@ -0,0 +1,97 @@
+import numpy as np
+from baseline.utils import import_user_module
+
+class RemoteTFModel(object):
```
dpressel: Naming should be consistent with other places, `RemoteModelTensorFlow`

python/baseline/model.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 from baseline.utils import (
-    export, optional_params, listify, register
+    export, optional_params, listify, import_user_module, register
```
dpressel: superfluous

python/baseline/services.py
```diff
@@ -41,24 +45,75 @@ def get_labels(self):
...
+        """
+        assets = read_json(os.path.join(directory, 'model.assets'))
+        model_name = assets['metadata']['exported_model']
+        labels = read_json(os.path.join(directory, model_name) + '.labels')
+        lengths_key = assets.get('lengths_key', None)
+        inputs = assets.get('inputs', [])
+
+        model = None
+        if backend == 'tf':
+            model = RemoteTFModel(remote,
```
dpressel: This didnt get renamed

PR #163, opened 2018-10-23
Guarantee offsets for <PAD>, <GO>, <EOS>, <UNK>

dpressel:
This change guarantees special tokens <PAD>, <GO>, <EOS> and <UNK> in that order.
It creates an `Offsets` object in `baseline/utils.py` that provides the offsets and their values.
Also refactored `baseline/w2v.py` to generically preset these vectors and removed quirky
constructor params from seq2seq modules to simplify things.
Additionally, we ensure that the first 3 Offsets are also in the label vectorizer for tagging

### DIFF
--- a/python/addons/tagger_elmo.py
+++ b/python/addons/tagger_elmo.py
@@ -9,6 +9,7 @@ from baseline.model import TaggerModel, create_tagger_model, load_tagger_model
 import tensorflow as tf
 import tensorflow_hub as hub
 import numpy as np
+from baseline.utils import Offsets
 
 
 class RNNTaggerModelELMoModel(TaggerModel):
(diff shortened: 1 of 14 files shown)


blester125:
I don't know if it is useful but you can make offsets iterable and star unpackable like this

```python
from six import with_metaclass
from collections import Mapping

class Unpackable(type):
    def __iter__(self):
        for x in self.VALUES:
            yield x

    def __getitem__(self, item):
        item = item[1:-1]
        return getattr(self, item)

    def __len__(self):
        return len(self.VALUES)

class UnpackableMapping(Unpackable, Mapping): pass

class Offsets(with_metaclass(UnpackableMapping)):
    PAD, GO, EOS, UNK, OFFSET = range(0, 5)
    VALUES = ['<PAD>', '<GO>', '<EOS>', '<UNK>']
```

```pycon
>>> from off import Offsets
>>> for x in Offsets:
...     print(x)
... 
<PAD>
<GO>
<EOS>
<UNK>
>>> dict(**Offsets)
{'<PAD>': 0, '<GO>': 1, '<EOS>': 2, '<UNK>': 3}
>>> print(*Offsets)
<PAD> <GO> <EOS> <UNK>
```

dpressel:
This is a good idea and it seems like it will be compatible with my change in this PR.  For now, I will merge the PR, and we can do this separately when there is a bit more time.

PR #167, opened 2018-10-24
Share more code across services

blester125:


python/baseline/services.py
```diff
@@ -25,10 +25,9 @@
 __all__ = []
 exporter = export(__all__)
 
-
-@exporter
-class ClassifierService(object):
-
+class Service(object):
+    _name = None
```
dpressel: we use task_name elsewhere

PR #171, opened 2018-11-02
Added the option of a dilated convolution to parallel_conv and some

blester125:
tests for it


dpressel:
Will likely move this into v1 since we are close to release there

dpressel:
Closing for now, change would have to be made across frameworks and not sure its worth it.  If you feel different, please update with implementations for each F/W and I would reconsider merging

PR #178, opened 2018-11-08
Feature/nstep

blester125:
This PR adds the ability to have nstep reporting for each task and framework. This is used mostly for tensorboard reporting.

This adds reporting hooks that are step based and and those that are epoch based. They just ignore ones that they aren't set to use.

This also adds a new logging reporter that logs to stdout using the root logger. This allows for long running model (seq2seq and lm) to log to screen without polluting the reporting log.

### DIFF
--- a/python/baseline/tf/seq2seq/train.py
+++ b/python/baseline/tf/seq2seq/train.py
@@ -6,8 +6,6 @@ import tensorflow as tf
 from baseline.tf.optz import optimizer
 from baseline.utils import listify, get_model_file
 from baseline.train import Trainer, create_trainer, register_trainer, register_training_func
-import time
-import os
 
 
 @register_trainer(task='seq2seq', name='default')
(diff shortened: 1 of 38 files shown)


python/baseline/tf/seq2seq/train.py
```diff
@@ -74,24 +96,25 @@ def test(self, vs, reporting_fns, phase='Valid'):
...
+        total_toks = 0
         metrics = {}
 
         start = time.time()
         for batch_dict in vs:
 
             feed_dict = self.model.make_input(batch_dict)
-            vals = self.sess.run([self.test_loss], feed_dict=feed_dict)
-            lossv = vals[0]
+            vals = self.model.sess.run(fetches, feed_dict)
```
dpressel: `fetches` is not declared in test so this code doesnt work

python/baseline/tf/seq2seq/train.py
```diff
@@ -34,37 +41,52 @@ def recover_last_checkpoint(self):
...
+
     def train(self, ts, reporting_fns):
-        total_loss = 0
-        steps = 0
-        metrics = {}
-        duration = 0
+        epoch_loss = 0
+        epoch_toks = 0
+
+        fetches = {
```
dpressel: This code doesnt appear to be used and should be removed

PR #181, opened 2018-11-13
Fixed pytorch types

blester125:
Changing the vectorizers to force `np.int32` caused pytorch to have some problems. Pytorch wants LongTensors rather than IntTensors.

This change casts the 'y' target values to LongTensors during `make_input`.

Other inputs are cast to Longs when they are used in the embeddings layer. Casting a `[100 x 100]` tensor on GPU takes `14.4 micro_s +/- 231 ns` This doesn't seem like a huge performance hit but it would faster to do the cast once during `make_input`. However because we don't know which features are indices to embedding tables and which could be continuous pass through features I don't think we can just cast everything in the loop.

### DIFF
--- a/python/baseline/pytorch/classify/model.py
+++ b/python/baseline/pytorch/classify/model.py
@@ -66,7 +66,7 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
 
         y = batch_dict.get('y')
         if y is not None:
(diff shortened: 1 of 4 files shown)


dpressel:
I think this is solved by https://github.com/dpressel/baseline/pull/182

PR #198, opened 2018-11-23
Feature/dynet attn

blester125:
This PR updates Dynet attention to include more kinds of attention (bahdanau, bilinear, dot, and scaled dot product).

This also updates the attention combination. Rather than just returning the weighted average of the encoder vectors it also concatenates the decoder state and down projects it to the right size like the pytorch version does.

### DIFF
--- a/python/baseline/dy/dynety.py
+++ b/python/baseline/dy/dynety.py
@@ -1,3 +1,4 @@
+import math
 from itertools import chain
 import numpy as np
 import dynet as dy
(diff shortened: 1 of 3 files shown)


python/baseline/dy/dynety.py
```diff
@@ -387,45 +397,163 @@ def conv(input_, _=None):
...
+        if mask is not None:
+            attn_scores = dy.cmult(attention_scores, mask[0]) + (mask[1] * dy.scalarInput(-1e9))
+        return dy.softmax(attn_scores)  # ((T, 1), B)
+
+    def _combine(self, attn, query):
+        comb = super(ScaledDotProductAttention, self)._combine(attn, query)
+        return dy.tanh(comb)
+
+
+class LoungAttention(Attention):
```
dpressel: Spelling is wrong

PR #199, opened 2018-11-26
support HTTP remote TF service

dpressel:
When HTTP is on, we dont have any TF or gRPC deps

### DIFF
--- a/python/baseline/tf/remote.py
+++ b/python/baseline/tf/remote.py
@@ -1,6 +1,85 @@
 import numpy as np
 from baseline.utils import import_user_module
-import tensorflow as tf
+
+
+class RemoteModelTensorFlowREST(object):
+
+    def __init__(self, remote, name, signature, labels=None, beam=None, lengths_key=None, inputs=[]):
+
+        self.remote = remote
+        self.name = name
+        self.signature = signature
+        self.lengths_key = lengths_key
+        self.input_keys = set(inputs)
+        self.beam = beam
+        self.labels = labels
+
+    def get_labels(self):
+        return self.labels
+
+    def predict(self, examples):
+        import requests
+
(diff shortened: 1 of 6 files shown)


python/baseline/tf/remote.py
```diff
@@ -1,6 +1,85 @@
...
+            raise ValueError("should have keys: " + ",".join(self.input_keys))
+
+        url = '{}/v1/models/{}/versions/1:predict'.format(self.remote, self.name)
+        request = self.create_request(examples)
+        outcomes_list = requests.post(url, json=request)
+        outcomes_list = outcomes_list.json()['outputs']
+        outcomes_list = self.deserialize_response(examples, outcomes_list)
+        return outcomes_list
+
+    def deserialize_response(self, examples, predict_response):
```
mtbarta: there could be a helper method here shared between rest and grpc for assembling classes and scores.

python/mead/config/twpos.json
```diff
@@ -48,7 +48,7 @@
 		"dsz": 30,
 		"wsz": 30,
 		"type": "char-conv",
-		"cfiltsz": [1,2,3,4,5,7]
+		"cfiltsz": [3]
```
mtbarta: why is this change in this PR?

python/baseline/tf/remote.py
dpressel: maybe, but I dont think it will really eliminate any code, and it may cause the code to be harder to read

python/mead/config/twpos.json
dpressel: Agree this doesnt necessarily belong in the PR, it turns out that all these filters dont seem to really help performance on the current baseline model

PR #212, opened 2018-12-03
WIP: Feature/save state tf

dpressel:
This modification creates `.state` field which is attached onto the model in a `.create()` call.  This set of fields is serialized on `.save()` calls and the file is reloaded during `.load()` to determine `kwargs` to pass into `.create()` to reconstruct the model.  This eliminates the need to write the TF graph out, as all of the components necessary to reconstruct exist in the `.state` file.  This also simplifies the path for reloading since it can just delegate to `.create()` under the hood.

Additionally, this change may make it possible to further simplify the exporters, which can use the `.state` file to build the graph, rather than the mead config.


python/baseline/tf/classify/model.py
```diff
@@ -13,7 +13,7 @@
 from tensorflow.contrib.layers import fully_connected
 import os
 import copy
-
+import copy
```
blester125: Copy is imported twice

python/baseline/tf/lm/model.py
```diff
@@ -222,6 +199,18 @@ def predict(self, batch_dict):
         step_softmax = self.sess.run(self.probs, feed_dict)
         return step_softmax
 
+    def _create_model_state(self, **kwargs):
+        embeddings_info = {}
+        for k, v in self.embeddings.items():
+            embeddings_info[k] = v.__class__.__name__
+        sess = kwargs.pop('sess', tf.Session())
```
blester125: How does reloading wrok with multigpu? It just reloads a single model or will it be able to reload and redistribute to gpus?

Do you also need to pop the `'id'` key here?

python/baseline/tf/classify/model.py
```diff
@@ -281,62 +269,48 @@ def load(cls, basename, **kwargs):
...
-
-        model.embeddings = dict()
-        for key, class_name in state['embeddings'].items():
+        state['sess'] = kwargs.pop('sess', tf.Session())
+        state['model_type'] = kwargs.get('model_type', 'default')
+        embeddings = dict()
+        embeddings_dict = state.pop('embeddings')
+        for key, class_name in embeddings_dict.items():
             md = read_json('{}-{}-md.json'.format(basename, key))
             embed_args = dict({'vsz': md['vsz'], 'dsz': md['dsz']})
```
blester125: Are embedding objects also getting this save kwargs treatment? If so shouldn't the embed args be the whole contents of the md file?

dpressel:
This item is on hold until we figure out the best way to determine which kwargs should be persisted and which should not.  For example, if we are injecting our inputs into the model, this paradigm doesnt work.  The idea of registering each property on the class seems onerous and very un-user friendly, but if you dont do that, you have to figure out some general purpose way of blacklisting particular kwargs, which seems equally unappealing

PR #222, opened 2019-01-07
Tidying tf tagger

mtbarta:
This PR reorganizes tf/TaggerModelBase.

* brought the init method to the top of the class
* moved properties underneath init
* save method lives closer to its children methods


dpressel:
Will close this for now, as things are moving anyway in the branch

PR #223, opened 2019-01-08
Updated Beam Searches and Batched pytorch one.

blester125:
This PR is an update of decoding for seq2seq models in pytorch as well as a few bug fixes for other frameworks.

Bugfixes:
 * Pytorch and Dynet are updated to do greedy decoding on their dev sets for Bleu score instead of teacher forcing. (TF has not be updated)
 * Pytorch and Dynet score masking is updated so that once a beam is finished all probability is placed on the EOS token, preventing a bug where the same beam is selected K times.
 * Fixed an error in pytorch Bahdanau attnetion where the mask was applied but the result was not used causing errors when used on a batch.

Small Tweaks:
 * Moved the normalize backend code into baseline to allow for use in services without a dependency on mead. This lets you say `python ed-text.py --backend pyt`
 * Pytorch `seq2seq.predict` can now be told if the input needs to go through `make_input`.
 * Pytorch `seq2seq.make_input` (for seq2seq) can now return the permutation index so that the original order can be restored.
 * `services.EncoderDecoderService` now uses the `lookup_sentence` function to translate from indices to words.

Updates:
 * Batched Beam search for Pytorch! (5m32 on master for dev -> 20s)
 * Abstracted beam search into a function that takes 3 main callables, ones that initialize the state for the search, one that runs a single step, and a final one that updated the state based on the best beams. I would like some input on if others like how this is implemented.
 * A discussion to have is the way i am tracking done, I am using a single length where a 0 means it hasn't ended. Would it be better to have a length that is always updated and a separate done tracker?
 * Adds the ability to have a length penalty based on the gnmt paper. This was able to give a bit more than a point boost to Bleu scores.


blester125:
Updated to remove the Transformer for now

PR #227, opened 2019-01-14
support an output from the seq2seq up to beam len

dpressel:
- If K=1, this will produce previous output
- When K > 1, will create an additional list sequence for each beam output

### DIFF
--- a/python/baseline/services.py
+++ b/python/baseline/services.py
@@ -372,7 +372,7 @@ class EncoderDecoderService(Service):
         kwargs['beam'] = kwargs.get('beam', 10)
         return super(EncoderDecoderService, cls).load(bundle, **kwargs)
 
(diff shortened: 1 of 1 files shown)


python/baseline/services.py
```diff
@@ -408,16 +408,25 @@ def predict(self, tokens, **kwargs):
...
-            results += [out]
+        B = len(outcomes)
+        for i in range(B):
+            N = len(outcomes[i])
+            n_best_result = []
+            for n in range(min(K, N)):
+                n_best = outcomes[i][n]
+                out = []
+                T = len(n_best)
+                for j in range(T):
```
blester125: We might as well use the lookup_sentence utility here

dpressel: looks like you have a modification to services on deck, so marked as resolved

PR #228, opened 2019-01-14
Make Services easier to use.

blester125:
When building tools that use the service objects to facilitate model reloading they become hard to work with when doing much more than the normal prediction. This PR splits the data preprocessing into helper functions that can be used independently.


blester125:
Updated to fit with the new beam output changes

PR #229, opened 2019-01-24
Preproc exporter

sagnik:
preproc exporter for classification and tagging

-  exports a word convolutional model for classification and word + char conv BLSTM model for tagging. 
-  preproc code has been moved to the add on area.
-  services code has been updated to take preproc argument: if the preproc argument is true, no vectorization happens.
-  tests: 1. `<baseline-dir>/tests/test-export-classify.sh`, 2. `<baseline-dir>/tests/test-export-tagger.sh` :
    -   saved models and test files are downloaded (sst2 and conll)
    -   model is _loaded_ and the file is classified (tagged).
    -   model is _exported w/o preproc_ and the file is classified (tagged).
    -   model is _exported w preproc_ and the file is classified (tagged). 
    -   the outputs are compared for equality.


python/mead/config/conll-single-tf.json
```diff
@@ -2,6 +2,7 @@
     "task": "tagger",
     "batchsz": 10,
     "conll_output": "conll-iobes-results.conll",
+    "modules":["preproc-exporters"],
```
dpressel: I wonder if we can think of a better way to make this an export-only concern

api-examples/classify-text.py
```diff
@@ -1,12 +1,15 @@
...
+from baseline.utils import str2bool
+
 parser = argparse.ArgumentParser(description='Classify text with a model')
-parser.add_argument('--model', help='A classifier model', required=True, type=str)
+parser.add_argument('--model', help='A directory containing classifier model files', required=True, type=str)
 parser.add_argument('--text', help='raw value', type=str)
 parser.add_argument('--backend', help='backend', default='tf')
 parser.add_argument('--remote', help='(optional) remote endpoint', type=str) # localhost:8500
 parser.add_argument('--name', help='(optional) service name', type=str)
+parser.add_argument('--preproc', help='(optional) set to true if want to use preproc', type=str2bool, default=False)
```
dpressel: Shouldnt this be an `exporter_type` not a `preproc` option?  Right now it seems really tightly coupled to a very specific addon, the other way would be more general

python/baseline/services.py
```diff
@@ -173,18 +178,19 @@ def task_name(cls):
...
         self.set_vectorizer_lens(mxlen, mxwlen)
         examples = self.vectorize(token_seq)
-
+        if preproc:
+            examples['tokens'] = [" ".join(x) for x in tokens]
         outcomes_list = self.model.predict(examples)
         results = []
         for outcomes in outcomes_list:
-            results += [list(map(lambda x: (x[0], x[1].item()), sorted(outcomes, key=lambda tup: tup[1], reverse=True)))]
+            results += [list(map(lambda x: (x[0], x[1]), sorted(outcomes, key=lambda tup: tup[1], reverse=True)))]
```
dpressel: was there a reason that we dont need `item()` here anymore?

python/addons/preprocessors.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Make a vocab list
+        vocab_list = [''] * (len(vocab) + 1)
+
+        for v, i in vocab.items():
+            vocab_list[i] = v
+
+        tok2index = tf.contrib.lookup.index_table_from_tensor(
+            tf.constant(vocab_list),
+            default_value=3, # This is since we have moved all <UNK>s to 3
```
blester125: This should use `Offsets.UNK`

python/addons/preproc-exporters.py
```diff
@@ -0,0 +1,64 @@
...
+@register_exporter(task='tagger', name='preproc')
+class TaggerTensorFlowPreProcExporter(TaggerTensorFlowExporter):
+
+    def __init__(self, task):
+        super(TaggerTensorFlowPreProcExporter, self).__init__(task)
+
+    def _create_rpc_call(self, sess, model_file):
+        model_base_dir = os.path.split(model_file)[0]
+        pid = model_file.split("-")[-1]
+        features = ["word", "char"]
```
blester125: This makes it so that I need to have my features named `word` and `char`. Can these feature names be read from the config file so different/more features are allowed?

api-examples/classify-text.py
sagnik: `export.py` has an `exporter-type` option. this option is not tied to an addon. it just tells the `predict` methods to use/ not use preproc.

python/baseline/services.py
sagnik: I had a discussion with @blester125 about this. seems the pyt local models here returns a tensor whereas the remote returns a scalar. so a better idea would be to make the remote models behave the same as the local.

python/mead/config/conll-single-tf.json
sagnik: we can always have a separate config file with this change, will that work?

api-examples/classify-text.py
```diff
@@ -20,7 +23,9 @@
 else:
     texts = [args.text.split()]
 
-print(texts)
-
-m = bl.ClassifierService.load(args.model, backend=args.backend, remote=args.remote, name=args.name)
-print(m.predict(texts))
+m = bl.ClassifierService.load(args.model, backend=args.backend, remote=args.remote, name=args.name, preproc=args.preproc)
+for text, output in zip(texts, m.predict(texts, preproc=args.preproc)):
+    if type(output[0][0]) is not str:  # remote models return binary strings
```
sagnik: This should be handled by the model itself, as pointed out by @blester125

python/addons/preproc-exporters.py
sagnik: @dpressel WDYT?

sagnik: nvrmind. i changed the code.

python/addons/preprocessors.py
```diff
@@ -0,0 +1,175 @@
...
+        # sentence length <= mxlen
+        raw_post = tf.reduce_join(raw_tokens[:self.mxlen], separator=" ")
+        return raw_post
+
+    def create_word_vectors_from_post(self, raw_post):
+        # vocab has only lowercase words
+        word2index = self.indices['word']
+        split_chars = tf.string_split(tf.reshape(raw_post, [-1]), delimiter="").values
+        upchar_inds = self.upchars_lut.lookup(split_chars)
+        lc_raw_post = tf.reduce_join(tf.map_fn(lambda x: tf.cond(x[0] > 25,
```
blester125: Shouldn't this be conditional? there are times when people don't want to lowercase their input

python/addons/preprocessors.py
```diff
@@ -0,0 +1,176 @@
...
+        x = tf.sparse_tensor_to_dense(reshaped)
+        x = tf.contrib.framework.with_shape(shape, x)
+        return x
+
+
+class Token1DPreprocessorCreator(PreprocessorCreator):
+
+    def __init__(self, model_base_dir, pid, features, **kwargs):
+        super(Token1DPreprocessorCreator, self).__init__(model_base_dir, pid, features, **kwargs)
+        self.mxlen = self.vectorizers['word'].mxlen
```
blester125: This is still hard coded to `word` rather than being based on the `primary_key` idea from mead

blester125:
I trained a sst2 model with this feature set, it was able to be exported but when I tried to use it with `classify-text.py` I got the error below.

```json
  "features": [
    {
      "name": "word",
      "vectorizer": {
        "type": "token1d",
	"transform": "baseline.lowercase"
      },
      "embeddings": {
        "label": "w2v-gn"
      }
    },
    {
      "name": "char",
      "vectorizer": {
        "type": "char2d"
      },
      "embeddings": { "dsz": 30, "wsz": 30, "type": "char-conv" }
    }
  ],
```

```traceback
python classify-text.py --model ../models/client/models/1/ --remote localhost:8500 --name test --text "I hate this movie so much" --preproc true
../models/client/models/1/model.assets
../models/client/models/1/vocabs-char-5506.json
../models/client/models/1/vocabs-word-5506.json
Traceback (most recent call last):
  File "classify-text.py", line 27, in <module>
    for text, output in zip(texts, m.predict(texts, preproc=args.preproc)):
  File "/home/blester/dev/work/baseline/python/baseline/services.py", line 190, in predict
    outcomes_list = self.model.predict(examples)
  File "/home/blester/dev/work/baseline/python/baseline/remote.py", line 161, in predict
    outcomes_list = stub.Predict(request)
  File "/home/blester/anaconda3/envs/dl/lib/python3.6/site-packages/grpc/_channel.py", line 500, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/home/blester/anaconda3/envs/dl/lib/python3.6/site-packages/grpc/_channel.py", line 434, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, You must feed a value for placeholder tensor 'char' with dtype int32 and shape [?,?,?]
	 [[{{node char}} = Placeholder[_output_shapes=[[?,?,?]], dtype=DT_INT32, shape=[?,?,?], _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]])>
```

It seem that the way it is currently is coupled with the idea that the classifiers have one feature [(word)](https://github.com/dpressel/baseline/blob/50c7a692bc6fb668e78820ef1e244ffcfc1cd556/python/addons/preprocessors.py#L130) and the taggers have two [(word and char)](https://github.com/dpressel/baseline/blob/50c7a692bc6fb668e78820ef1e244ffcfc1cd556/python/addons/preprocessors.py#L163)

I would think that these would be dynamic based on the features and the vecotrizers to allow for things like word and character classifiers

python/addons/preprocessors.py
sagnik: > I trained a sst2 model with this feature set, it was able to be exported but when I tried to use it with `classify-text.py` I got the error below.
> 
> ```json
>   "features": [
>     {
>       "name": "word",
>       "vectorizer": {
>         "type": "token1d",
> 	"transform": "baseline.lowercase"
>       },
>       "embeddings": {
>         "label": "w2v-gn"
>       }
>     },
>     {
>       "name": "char",
>       "vectorizer": {
>         "type": "char2d"
>       },
>       "embeddings": { "dsz": 30, "wsz": 30, "type": "char-conv" }
>     }
>   ],
> ```
> 
> ```
> python classify-text.py --model ../models/client/models/1/ --remote localhost:8500 --name test --text "I hate this movie so much" --preproc true
> ../models/client/models/1/model.assets
> ../models/client/models/1/vocabs-char-5506.json
> ../models/client/models/1/vocabs-word-5506.json
> Traceback (most recent call last):
>   File "classify-text.py", line 27, in <module>
>     for text, output in zip(texts, m.predict(texts, preproc=args.preproc)):
>   File "/home/blester/dev/work/baseline/python/baseline/services.py", line 190, in predict
>     outcomes_list = self.model.predict(examples)
>   File "/home/blester/dev/work/baseline/python/baseline/remote.py", line 161, in predict
>     outcomes_list = stub.Predict(request)
>   File "/home/blester/anaconda3/envs/dl/lib/python3.6/site-packages/grpc/_channel.py", line 500, in __call__
>     return _end_unary_response_blocking(state, call, False, None)
>   File "/home/blester/anaconda3/envs/dl/lib/python3.6/site-packages/grpc/_channel.py", line 434, in _end_unary_response_blocking
>     raise _Rendezvous(state, None, None, deadline)
> grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.INVALID_ARGUMENT, You must feed a value for placeholder tensor 'char' with dtype int32 and shape [?,?,?]
> 	 [[{{node char}} = Placeholder[_output_shapes=[[?,?,?]], dtype=DT_INT32, shape=[?,?,?], _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]])>
> ```
> It seem that the way it is currently is coupled with the idea that the classifiers have one feature [(word)](https://github.com/dpressel/baseline/blob/50c7a692bc6fb668e78820ef1e244ffcfc1cd556/python/addons/preprocessors.py#L130) and the taggers have two [(word and char)](https://github.com/dpressel/baseline/blob/50c7a692bc6fb668e78820ef1e244ffcfc1cd556/python/addons/preprocessors.py#L163)
> 
> I would think that these would be dynamic based on the features and the vecotrizers to allow for things like word and character classifiers
> 
> I get a similar error when I have word and glove as features, The placeholder for for glove in not filled and GRPC complains about it.

i think the i was going with the idea that if you have a custom model you write a preproc exporter for that. but this can be generalized a bit i think.

sagnik: IOW preproc extensions are there to handle the transform fns. so if in your prev eg you had a transform fn for char then it won't be possible to generalize for that.

python/tests/test-export-classify.sh
```diff
@@ -0,0 +1,86 @@
+BASELINE_DIR=${HOME}/dev/work/baseline
+CLASSIFY_MODEL=sst2-25195.zip
+TEST_FILE=stsa.binary.test
+TEST_LOAD=$TEST_FILE.load
+TEST_SERVE=$TEST_FILE.serve
+TEST_SERVE_PREPROC=$TEST_FILE.serve_preproc
+SLEEP=5
+
+docker stop tfserving
```
blester125: Is this an alias on your system? I get this error

```
Error response from daemon: No such container: tfserving
Error: No such container: tfserving
```

The docs says the image is called tensorflow/serving

sagnik: ignore the error.

python/addons/preprocessors.py
sagnik: the code is updated to do this.

python/tests/test-export-tagger.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "exporting w/o preproc"
+MDIR=models/conll
+rm -rf $MDIR
+mkdir -p $MDIR
+mead-export --config $BASELINE_DIR/python/mead/config/conll.json --model $TAGGER_MODEL --is_remote false --output_dir $MDIR
+sleep $SLEEP
+
+echo "running tf serving"
+docker stop tfserving
```
blester125: `> /dev/null 2>&1` will suppress the error of not finding the container which will help clean up this output.

python/tests/test-export-tagger.sh
```diff
@@ -0,0 +1,87 @@
...
+    if [ "$DIFF" != "" ]
+    then
+        echo $TEST_LOAD does not match with $TEST_SERVE, exporting failed.
+        exit 1
+    fi
+
+DIFF=$(diff ${TEST_SERVE} ${TEST_SERVE_PREPROC})
+    if [ "$DIFF" != "" ]
+    then
+        echo $TEST_SERVE does not match with $TEST_SERVE_PREPROC, exporting failed.
```
blester125: There is a lot of output from theses scripts, it would be nice to have some colors on the fail/pass

```bash
RED='\033[0;31m'
GREEN='\033[0;32m'
END='\033[0m'

echo -e "${RED}You blew it${END}"
echo -e "${GREEN}You passed!${END}"
```

python/tests/test-export-tagger.sh
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+BASELINE_DIR=${HOME}/dev/work/baseline
```
blester125: Finding the baseline dir should probably be done by either back tracking from here or with `git rev-parse --show-toplevel` rather than relaying on it being at `~/dev/work`

python/tests/test-export-classify-preproc.sh
```diff
@@ -0,0 +1,14 @@
+BASELINE_DIR=${HOME}/dev/work/baseline
```
blester125: Can this file be deleted?

python/addons/preproc-exporters.py
```diff
@@ -0,0 +1,149 @@
...
+    def _create_rpc_call(self, sess, model_file):
+        model_base_dir = os.path.split(model_file)[0]
+        pid = model_file.split("-")[-1]
+        pc = PreProcessorController(model_base_dir, pid, self.task.config_params['features'])
+        model_params = self.task.config_params['model']
+        tf_example, preprocessed = pc.run()
+        for feature in preprocessed:
+            model_params[feature] = preprocessed[feature]
+        model, classes, values = self._create_model(sess, model_file)
+        sig_input = {'tokens': tf.saved_model.utils.build_tensor_info(tf_example[pc.FIELD_NAME])}
```
blester125: Shouldn't it be possible to have a lengths key for the classifier? LSTM models use it.

python/tests/test-export-classify.sh
```diff
@@ -0,0 +1,89 @@
+BASELINE_DIR=${HOME}/dev/work/baseline
```
blester125: There should probably be a version check for tensorflow here, My tf 1.8 install was failing these scripts.

```bash
python -c 'import tensorflow; from distutils.version import LooseVersion; import sys; i = 1 if LooseVersion(tensorflow.__version__) < LooseVersion("1.12.0") else 0; sys.exit(i)'
```

Then you can check the return value to see if you should abort

python/addons/preprocessors.py
```diff
@@ -0,0 +1,119 @@
...
+        # Reshape them out to the proper length
+        reshaped_words = tf.sparse_reshape(word_indices, shape=[-1])
+        return self.reshape_indices(reshaped_words, [self.mxlen])
+
+    def preproc(self, post_mappings):
+        # Split the input string, assuming that whitespace is splitter
+        # The client should perform any required tokenization for us and join on ' '
+        raw_post = post_mappings[self.FIELD_NAME]
+        raw_post = self.resize_sen(raw_post)
+        return self.create_word_vectors_from_post(raw_post)
```
blester125: Token1d should check the `self.vectorizer.transform_fn.__name__` do decide if it should lowercase or not.

If this function is not lowercase or identity it should probably raise an error that it can't export arbitrary functions.

sagnik: agree with the first part, the second i would think is the designer's responsibility.

blester125: By designer you mean who ever is writing a plugin? I think that makes sense but in this case the designer is you and I think for our use cases a safe guard like this makes sense, especially because it's not super easy to tell after exporting if it is lowercasing or not

python/addons/preproc-exporters.py
sagnik: added

python/baseline/services.py
```diff
@@ -266,10 +272,13 @@ def predict(self, tokens, **kwargs):
         :param tokens: (``list``) A list of tokens
 
         """
+        preproc = kwargs.get('preproc', False)
         label_field = kwargs.get('label', 'label')
         tokens_seq, mxlen, mxwlen = self.batch_input(tokens)
         self.set_vectorizer_lens(mxlen, mxwlen)
         examples = self.vectorize(tokens_seq)
+        if preproc:
+            examples['tokens'] = [" ".join(x) for x in tokens]
```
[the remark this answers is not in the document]
sagnik: This requires to be `tokens`, apparently `tokens_seq` is a dictionary?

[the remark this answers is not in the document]
sagnik: hmm i see where the problem is,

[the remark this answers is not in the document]
sagnik: changed, resolving.

python/mead/tasks.py
```diff
@@ -355,15 +355,21 @@ def _create_backend(self):
             backend.params = {'pc': _dynet.ParameterCollection(), 'batched': batched}
         elif backend.name == 'tf':
             # FIXME this should be registered as well!
-            from mead.tf.exporters import ClassifyTensorFlowExporter
-            backend.exporter = ClassifyTensorFlowExporter
+            exporter_type = kwargs.get('exporter_type', 'default')
```
blester125: If we had folders for `mead/pytorch` and `mead/dy` that are empty (because we don't support export yet) could we replace this with a `load_user_module('mead.{backend}')` in the `backend.load` call? This should load the registries that the `create_exporter` call in `mead-export` uses

sagnik: i dont think that's a necessary change. WDYT @dpressel

blester125: I don't think its necessary and can be merged without it but it seems like what will happen moving forward especially as pytorch gets more production features.

PR #240, opened 2019-02-13
Make preproc config param optional

sagnik:
in `v1`, the configs should not have a `preproc` section, but there were places in the code where we were assuming that section existed. this PR fixes that.

tested with one config. but do not merge it now, ~i will change all configs by removing the `preproc` section~. i will update the PR. 

As we decided to keep `preproc` i think the PR should be ok to merge, i am not changing the other configs to remove the preproc section. I have tested with one classification and one tagger config. worked for both.

There's another change: needed for tagger models to work when tokens are in the format `List[{'feature1': <>, 'feature2': <>}]`


blester125:
Idiomatic python uses `{}` to create an empty dictionary opposed to `dict()`

PR #243, opened 2019-02-15
tag-text accepts conll file + some other mods

sagnik:
1.  Due to some changes in the architecture, the exporter test script was failing because of an old model. new model trained, updated on dropbox.  
2.  tag-text takes conll file inputs + features (`feature:column number`). 
3.  tag-text also takes a `dict` argument called `grpc_feature_map` which defaults to `{tokens:text}`. If you have multiple features and the GRPC request needs to use different fields for these features, you should use this. eg: `{'tokens': 'text', 'ner': 'ner'}`.
4. for the same purpose, exporter can read a field called `exporter_field` from the config. Eg:
```
"features": [
        {
            "name": "word",
            "vectorizer": {
                "type": "dict1d",
                "fields": "text",
                "transform":"baseline.lowercase"
            },
            "embeddings": { "label": "glove-6B-100" },
            "exporter_field": "tokens"
        },
```   
5. exporter also takes an optional argument `return_labels`. This tells whether the exported model will return the class lables (if true) or indices. default value: true. this is stored in `assets` and is read by the client.
6. if your model had a mxlen and mxwlen defined in your vectorizers, `batch_input` methods did not care about them. this is rectified. although i am not sure about s2s models, so let me know what to be done there.

**EDIT**

Removed WIP status


api-examples/tag-text.py
```diff
@@ -1,29 +1,62 @@
...
 parser.add_argument('--text', help='raw value', type=str)
+parser.add_argument('--conll', help='is file type conll?', type=str2bool, default=False)
+parser.add_argument('--features', help='(optional) features in the format feature_name:index (column # in conll) or '
+                                       'just feature names (assumed sequential)', default=[], nargs='+')
 parser.add_argument('--backend', help='backend', default='tf')
 parser.add_argument('--remote', help='(optional) remote endpoint', type=str) # localhost:8500
 parser.add_argument('--name', help='(optional) signature name', type=str)
 parser.add_argument('--preproc', help='(optional) where to perform preprocessing', choices={'client', 'server'}, default='client')
-
+parser.add_argument('--grpc_feature_map', help='mapping between features and the fields in the grpc request,'
```
dpressel: This seems unnecessarily complex for an end user.  They shouldn’t have to define a JSON object on the command line

sagnik: for `features`, you know the indices of the json are 0,1 ... That is not true for this argument.

python/baseline/services.py
```diff
@@ -122,7 +140,8 @@ def load(cls, bundle, **kwargs):
         name = kwargs.get("name", None)
         if remote:
             beam = kwargs.get('beam', 10)
-            model = Service._create_remote_model(directory, be, remote, name, cls.signature_name(), beam, preproc=kwargs.get('preproc', False))
+            model = Service._create_remote_model(directory, be, remote, name, cls.signature_name(), beam,
+                                                 preproc=kwargs.get('preproc', False))
```
blester125: This seems like it should default to `client` like the `kwargs.get('preproc', 'client')` does

python/baseline/services.py
```diff
@@ -273,21 +328,42 @@ def predict(self, tokens, **kwargs):
...
         outcomes = self.model.predict(examples)
+
         outputs = []
+        if hasattr(self.model, 'return_labels'):
+            return_labels = self.model.return_labels
+        else:
+            return_labels = False  # keeping the default tagger behavior
+
+        if not return_labels:
+            label_vocab = revlut(self.get_labels())
```
blester125: This will involve iterating though the label space on every call to predict. It would be nice to cache results or move this to where it doesn't happen all the time.

The same applies to the other Services

python/mead/export.py
```diff
@@ -36,8 +48,10 @@ def main():
 
     task = mead.Task.get_task_specific(task_name, args.logging, args.settings)
     task.read_config(config_params, args.datasets, exporter_type=args.exporter_type)
-
-    exporter = create_exporter(task, args.exporter_type)
+    feature_exporter_field_map = create_feature_exporter_field_map(config_params['features'])
+    # TODO: we should write out return_labels to model.assets and GRPC deserializers should take that info
```
blester125: This is already done

python/baseline/services.py
```diff
@@ -273,21 +328,42 @@ def predict(self, tokens, **kwargs):
...
+        # TODO: unlike classify, here we allow vectorizers even for preproc=server to get `word_lengths`.
+        # vectorizers should not be available when preproc=server.
         examples = self.vectorize(tokens_seq)
         if preproc == 'server':
-            examples['tokens'] = [" ".join([y['text'] for y in x]) for x in tokens_seq]
+            unfeaturized_examples = {}
+            for exporter_field in exporter_field_feature_map:
+                unfeaturized_examples[exporter_field] = [" ".join([y[exporter_field_feature_map[exporter_field]]
+                                                                   for y in x]) for x in tokens_seq]
+            unfeaturized_examples['word_lengths'] = examples['word_lengths']
```
blester125: This seems like it should check for some key to use for lengths rather then just defaulting `word_lengths`, I could be wrong though

python/mead/export.py
```diff
@@ -6,6 +6,17 @@
 from mead.exporters import create_exporter
 from baseline.utils import str2bool
 
+
+def create_feature_exporter_field_map(feature_section, default_exporter_field='tokens'):
```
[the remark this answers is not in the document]
sagnik: not sure about this comment. for preproc=client exporters `feature_exporter_field_map` field is never used.

python/baseline/services.py
sagnik: right. resolved.

python/mead/export.py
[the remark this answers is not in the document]
blester125: My bad, I missed that. This seems good then

python/baseline/services.py
sagnik: i think this was the default behavior in taggerservice predict and it just returns a previously stored dict. not sure if we would save a lot by changing the behavior.

python/baseline/services.py
sagnik: yes we could i think

api-examples/tag-text.py
```diff
@@ -1,28 +1,72 @@
...
+                text = line.strip().split()
+                texts += [text]
 
 else:
     texts = [args.text.split()]
 
 m = bl.TaggerService.load(args.model, backend=args.backend, remote=args.remote, name=args.name, preproc=args.preproc)
-for sen in m.predict(texts, preproc=args.preproc):
+for sen in m.predict(texts, preproc=args.preproc,
+                     exporter_field_feature_map=create_exporter_field_feature_map(args.export_mapping)):
```
dpressel: Rename to same as program arg

PR #255, opened 2019-03-11
Feature/pytorch export

blester125:
This PR introduces the current pytorch exporting functions.

 * It creates pytorch exporters that handle classifiers and taggers
 * It creates a Torchscript viterbi decoder to use for tagger
 * It moves common exporting functions into `mead.utils`
 * It creates sub files for dy and keras exporting so that the exporters can be loaded by the mead tasks backend object the same way the embeddings are
 * It creates example c++ pytorch files that can read from the exported bundle and perform inference with that model on command line input
 * It removes things that would try to trigger a multigpu create when exporting
 * It updates pytorch loading to give an optional `device` argument. This lets use reload a model that was saved on gpu on the cpu.
 * It converts the length argument to `pack_padded_sequence` from a list to a tensor (needed for exporting)
 * It updates pytorch models to also save out their labels when saving (needed for remote models)
 * It refactors remote models into baseclasses that are separated by transport protocol and moves backend specific implementations into `baseline.backend.remote`. It introduces the registry idea used elsewhere to create the specific remote models.
 * It adds an http server preproc remote model for tensorflow


api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+        word.begin(),
+        word.end(),
+        dst.begin(),
+        ::tolower
+    );
+}
+
+
+void find_vocab_file(const std::string& bundle, const std::string& vocab, std::string& dst) {
+    glob_t glob_result;
```
dpressel: not a huge deal, but glob is POSIX, not portable

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+    file_name << bundle << "/vocabs-" << vocab << "-*.json";
+    glob(file_name.str().c_str(), GLOB_TILDE, NULL, &glob_result);
+    dst = glob_result.gl_pathv[0];
+}
+
+
+int main(int argc, char** argv) {
+    if (argc < 3) {
+        std::cerr << "usage: classify-text export_bundle tokens...\n";
+        return -1;
```
dpressel: more common to use exit(1) on errors

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    std::string bundle(argv[1]);
+    std::vector<std::string> args(argv + 2, argv + argc);
+
+    std::stringstream model_file;
+    model_file << bundle << "/model.pt";
+    std::shared_ptr<torch::jit::script::Module> model = torch::jit::load(model_file.str());
```
dpressel: wouldnt `std::unique_ptr` work here?  Im not seeing an obvious need to have a smart ptr

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+using json = nlohmann::json;
+
+#include <iostream>
+#include <fstream>
+#include <memory>
+#include <string>
+#include <algorithm>
+#include <glob.h>
+
+#define UNK 3
```
dpressel: use `const long` not a preproc def.

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+            }
+            for (int j = 0; j < args_sz; j++) {
+                // Build it as a single long vector and reshape the tensor
+                int curr_sz = args[j].size();
+                for (int k = 0; k < mxwlen; k++) {
+                    if (k < curr_sz) {
+                        std::string c(1, args[j][k]);
+                        auto idx = vocabs[order[i]].find(c);
+                        if (idx != vocabs[order[i]].end()) {
+                            feature.push_back((long)vocabs[order[i]][c]);
```
dpressel: when you change to a const above, you will be able to eliminate these casts

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+                        } else {
+                            feature.push_back((long)UNK);
+                        }
+                    } else {
+                        feature.push_back((long)PAD);
+                    }
+                }
+            }
+            torch::Tensor f = torch::tensor(feature, torch::dtype(torch::kInt64));
```
dpressel: auto?

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::vector<torch::jit::IValue> features;
+
+    std::cout << "Building Features.\n";
+    int mxwlen = 0;
+    int args_sz = args.size();
+    for (int i = 0, sz = order.size(); i < sz; i++) {
+        std::vector<long> feature;
+        std::string feat = order[i];
+        // Create character based features
+        if (feat.compare("char") == 0) {
```
dpressel: is there a reason you are not just using `feat == "char"`?

api-examples/pytorch_cpp/classify-text.cpp
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+            torch::Tensor f = torch::tensor(feature, torch::dtype(torch::kInt64));
+            f = f.unsqueeze(0);
+            features.push_back(f);
+        }
+    }
+    input.push_back(torch::jit::Tuple::create(features));
+
+    torch::Tensor lengths = torch::full({1}, (int)args.size(), torch::dtype(torch::kInt64));
```
dpressel: auto?

api-examples/pytorch_cpp/classify-text.cpp
blester125: I added a mention of this in the README

api-examples/pytorch_cpp/classify-text.cpp
blester125: `std::shared_ptr` is what it was used in example code in the pytorch tutorial. I tried using `std::unique_ptr` and it wouldn't compile.

dpressel:
Please resolve conflicts and Ill merge

PR #260, opened 2019-03-15
remote.py replace requests lib with httplib and urllib

kiennguyen94:
`requests` takes up to 300ms to import. By simply replacing it with `httplib` and `urllib`, we are free from this run time deficit.

### DIFF
--- a/python/baseline/remote.py
+++ b/python/baseline/remote.py
@@ -1,5 +1,7 @@
 import numpy as np
+import json
 from baseline.utils import import_user_module
+from six.moves.http_client import HTTPConnection
 
 
 class RemoteModelTensorFlowREST(object):
@@ -19,6 +21,8 @@ class RemoteModelTensorFlowREST(object):
         :param version: The model version (defaults to None)
         """
         self.remote = remote
+        if len(self.remote.split(':')) != 3:
+            raise ValueError("remote has to have the form <host_name>:<port>")
         self.name = name
         self.signature = signature
         self.lengths_key = lengths_key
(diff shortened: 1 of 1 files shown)


python/baseline/remote.py
```diff
@@ -40,17 +42,24 @@ def predict(self, examples):
         :param examples: The input examples
         :return: The outcomes
         """
-        import requests
+        import httplib, urllib
```
blester125: The `urllib` import isn't needed and the import can be moved to the top of the file

python/baseline/remote.py
```diff
@@ -40,17 +42,24 @@ def predict(self, examples):
...
+        import json
```
blester125: The `requests` import was inside to make the dependency optional. Given that `json` is in the standard library we can move this import to the top of the file

python/baseline/remote.py
```diff
@@ -40,17 +42,24 @@ def predict(self, examples):
...
         v_str = '/versions/{}'.format(self.version) if self.version is not None else ''
         url = '{}/v1/models/{}{}:predict'.format(self.remote, self.name, v_str)
+        path = '/v1/models/{}{}:predict'.format(self.name, v_str)
         request = self.create_request(examples)
-        outcomes_list = requests.post(url, json=request)
-        outcomes_list = outcomes_list.json()
+        headers = {'Content-type': 'application/json' }
+        
+        _, hostname, port = self.remote.split(':')
+        conn = httplib.HTTPConnection(hostname.replace('//', ''), port)
```
blester125:
```suggestion
        conn = HTTPConnection(hostname.replace('//', ''), port)
```

This will make the change to use six easier

python/baseline/remote.py
blester125: `httplib` doesn't exist in python3. We use six for python2/3 stuff.

We need this line `from six.moves.http_client import  HTTPConnection` to make it work on 2 and 3

python/baseline/remote.py
```diff
@@ -40,17 +42,24 @@ def predict(self, examples):
...
-        import requests
+        import httplib, urllib
+        import json
 
         valid_example = all(k in examples for k in self.input_keys)
         if not valid_example:
             raise ValueError("should have keys: " + ",".join(self.input_keys))
 
         v_str = '/versions/{}'.format(self.version) if self.version is not None else ''
         url = '{}/v1/models/{}{}:predict'.format(self.remote, self.name, v_str)
```
blester125: It looks like `url` this can be removed now that it isn't used.

python/baseline/remote.py
```diff
@@ -19,6 +19,8 @@ def __init__(self, remote, name, signature, labels=None, beam=None, lengths_key=
         :param version: The model version (defaults to None)
         """
         self.remote = remote
+        if (len(self.remote.split(':')) != 3):
```
dpressel: can remove outer parens

python/baseline/remote.py
dpressel: still please move import up to the top of the file

python/baseline/remote.py
dpressel: this looks like it was moved to the top, but it never should be imported AFAICT

dpressel:
Looks good.  Thanks, Kien!

PR #262, opened 2019-03-18
updated to support `device` param

dpressel:
this change makes it possible to put a TF model on the CPU for inference inside of `services`.
Previously, the caller was responsible for this

### DIFF
--- a/api-examples/classify-text.py
+++ b/api-examples/classify-text.py
@@ -8,6 +8,7 @@ parser.add_argument('--text', help='raw value', type=str)
 parser.add_argument('--backend', help='backend', default='tf')
 parser.add_argument('--remote', help='(optional) remote endpoint', type=str) # localhost:8500
 parser.add_argument('--name', help='(optional) service name', type=str)
+parser.add_argument('--device', help='device', default='default')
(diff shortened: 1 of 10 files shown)


blester125:
When the pytorch export branch is eventually merged (or if we want to backport it) the pytorch models will be able  to use this `device` param too if we pass it in as a kwargs to the load_model_for call.

Do we want to move this code inside the load (to the models themselves) so there is a consistent interface once the pytorch changes are in?

dpressel:
my feeling is that we can do this in a few steps.

1. get interface merged so we support at services tier
2. merge pytorch PR
3. update (possibly at the model tier, as you suggest above) so that both frameworks support

dpressel:
I was able to get started on 3 and clean up 1 by creating a device wrapper decorator to wrap the `load()` inside the models.  This allows the service code to remain as before, and the model implementations barely change at all (they have to change a tiny bit, since the session being used wasnt correct).  The new change doesnt have any negative impact when the device is not specified.

The new decorator handles the cross-cutting concern of determining from the kwargs whether to create soft placement and a separate graph.

PR #263, opened 2019-03-18
Feature/iobes multi feature

sagnik:


python/baseline/utils.py
```diff
@@ -683,37 +683,38 @@ def convert_bio_to_iobes(ifile, ofile):
...
+                    next_tokens = lines[i+1].split()
+                    if len(next_tokens) > 1:
+                         next_tag = next_tokens[-1]
+                    else:
+                        next_tag = None
+
+                # Nothing to do for label == 'O'
+                if label == 'O':
                     updated_label = label
-                else:
```
sagnik: i am not sure where this change is coming from, i cant find this line in master: https://github.com/dpressel/baseline/blob/master/python/baseline/utils.py#L699 and that would make sense.

sagnik:
I am in favor of closing this: @blester125 's PR subsumes this I think?

PR #267, opened 2019-03-19
allow versioning in grpc TF

sagnik:
allow clients to access specific versions of the servable. tested internally.

### DIFF
--- a/python/baseline/remote.py
+++ b/python/baseline/remote.py
@@ -140,7 +140,7 @@ class RemoteModelTensorFlowGRPC(object):
         :param beam: The beam width (defaults to None)
         :param lengths_key: Which key is used for the length of the input vector (defaults to None)
         :param inputs: The inputs (defaults to empty list)
-        :param version: The model version (defaults to 1)
+        :param version: The model version (defaults to None)
         :param return_labels: Whether the remote model returns class indices or the class labels directly. This depends
         on the `return_labels` parameter in exporters
         """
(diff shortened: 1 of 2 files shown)


python/baseline/remote.py
```diff
@@ -140,6 +140,7 @@ def __init__(self, remote, name, signature, labels=None, beam=None, lengths_key=
         :param beam: The beam width (defaults to None)
         :param lengths_key: Which key is used for the length of the input vector (defaults to None)
         :param inputs: The inputs (defaults to empty list)
+        :param version: The model version (defaults to 1)
```
dpressel: default to None for backwards compat?

dpressel: also b/c user shouldnt necessarily need to know what version

python/baseline/services.py
```diff
@@ -145,7 +145,7 @@ def load(cls, bundle, **kwargs):
             beam = kwargs.get('beam', 10)
             model = Service._create_remote_model(directory, be, remote, name, cls.signature_name(), beam,
                                                  preproc=kwargs.get('preproc', 'client'),
-                                                 version=kwargs.get('version', 1))
+                                                 version=kwargs.get('version', None))
```
dpressel: None is default, dont need to specify that

python/baseline/remote.py
dpressel: change comment

PR #277, opened 2019-03-28
added support for dated keys of form x:y:datestamp

dpressel:
A user should be able to specify a generic key inside the mead config that applies to the latest version of a dataset.

Assume that dates are used to specify the latest version of a dataset, and that keys are of the form:
```
<x>
<y>:<datestamp1>
<y>:<datestamp2>
```
where `<datestamp*>` is a parseable date expression and `2` is assumed to be more recent than `1` in this case.

- If a user requests `<x>` they should get back `<x>`, the exact match
- If a user requests `<y>:<datestamp1>` they should also get back the exact match
- If a user requests `<y>` they should get back `<y>:<datestamp2>``

### DIFF
--- a/python/mead/tasks.py
+++ b/python/mead/tasks.py
@@ -17,6 +17,7 @@ from mead.utils import (
     get_mead_settings,
     print_dataset_info,
     read_config_file_or_json,
+    get_dataset_from_key
 )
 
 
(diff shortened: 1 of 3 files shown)


blester125:
In the commit message it says `x:y:datestamp` is this a typo it looks like this adds name, timestamp labels to the dataset config, what would be the use case for `x:y:datestamp`?

dpressel:
the `x` and `y` defined in the commit message predate the PR and are not intended to indicate the same thing.  For the purposes of the PR message, that would be `x:datestamp` or `y:datestamp`.  It is possible to have multiple `:`s as indicated in the commit message though.  This would be a matter of style

dpressel:
I removed the external dependency on `dateutil.parser`

PR #279, opened 2019-04-01
Remote Embeddings Service

blester125:
This PR does the following:

* Most of the Servable embeddings code is all in `python/addons/create_servable_embeddings.py`
  * Creates a small tensorflow `ServableTensorflowEmbeddingsModel` This is a small wrapper around a dict of embeddings to expose methods like `make_input`, `save`, and `predict`
  * Creates an `ServableEmbeddingTask` that loads the embeddings and saves the model
  * Creates `ServableEmbeddingsExporter` that is used to serve the embeddings models (normal and preproc)
  * When run as a script this addon creates and exports the embedding servable. It also is able to create a local version (in a zip file)
 * Fixes a small bug where the type information wasn't discoverable when using services for models exported with server preproc when used with gRPC
 * Fixes a small bug when searching for reader params in the lm task
 * Creates an `EmbeddingsService` to use these servable embeddings
 * Creates a version of the LUT embeddings that can bypass the 2GB tensor size limit.
 * Enforces the fact that feature names cannot have a `"-"` in them
 * Creates an example program that can read a text file, embed the text and save to an `.npy`
 * Adds to the preproc exporter the ability to find mxlen in a batch rather than using the mxlen from the vectorizer. Without this change it would always return `[B, vectorizer.mxlen, D]` which would be very wasteful, now it returns `[B, batch.mxlen, D]`
   * This is not currently extended to mxwlen
   * This also supports a mxlen from the vectorizer where the if the batch mxlen is larger than the vectorizers mxlen then the response is cut to the vectorizers mxlen.

 * Thorns:
   * Because there isn't a task dir (e.g. `baseline/tf/classify`) for this new task (`servable_embeddings`) we need to allow the import of that dir need to be skipped (`None` check in `mead.tasks.Backend` and the `try` in `baseline.services.Service.load`).
   * There is a weird import in the EmbeddingService to import that add on. Should this whole Service be moved into the addon too? We don't have register for this and it seems like it would be harder to use if we did


api-examples/embed-text.py
```diff
@@ -0,0 +1,36 @@
+import os
+import argparse
+import numpy as np
+import baseline as bl
+
+parser = argparse.ArgumentParser(description='Classify text with a model')
```
sagnik: the descriptions should change?

blester125: Fixed

python/baseline/tf/embeddings.py
```diff
@@ -6,6 +6,11 @@
 from baseline.embeddings import register_embeddings
 from baseline.tf.tfy import pool_chars, get_shape_as_list, stacked_lstm, embed
 
+
+FLOAT32 = 4
+GB2 = 2000000000
```
dpressel: 1024*1024*1024*2

blester125: updated

PR #284, opened 2019-04-06
WIP: Remove Tensorflow dep in grpc remote models.

blester125:
This PR remove the dependency on tensorflow when using the remote grpc models.

 * The needed protos are added to the repo, Not sure if we need these in the future but there is an edited version of the `prediction_service.proto` because we only use the predict api so removing the other apis means fewer `_pb2.py` files are needed.
 * There is a script that can build these protos but again probably not needed
 * The `_pb2.py` files are committed.
 * Remote services are updated to use these to build tensor protos rather than using tensorflow for it.


dpressel:
This needs more testing before it can be merged

dpressel:
I am going to close this for now as we are moving toward a 2.0 release ASAP

PR #287, opened 2019-04-11
Feature/classify output

sagnik:
This allows the user to output the test results on the `classify` task, similar to conll files in tagger.


dpressel:
Change to reader signature will probably cause issues with previous programs.  It might be possible to offer a new set of methods which do an “extended” read, or make the change for 2.0 where there will already be major changes

dpressel:
Im going to close this for now

PR #294, opened 2019-04-23
ensure use of python 3 pip module in installation script

ryangallen:


dpressel:
We currently support both python 2 and python 3.  Right now we assume that the user has chosen one or the other and go with that.

PR #301, opened 2019-05-09
Feature/xpctl server

sagnik:
`xpctl` as a server-client framework. 

1.  A [swagger spec](https://github.com/sagnik/baseline/blob/feature/xpctl-server/python/xpctl/xpctl.yaml)  generates a server and client.

2.  The spec defines two classes: `Experiment` and `ExperimentAggregate`. The `Experiment` class is self-explanatory: it has all info related to an experiment. All train_events, dev_events and test_events results are available to the `Experiment` class. 

3. `ExperimentAggregate` is generated by first grouping a set of Experiment objects by a reduction dimension and then running aggregate functions (mean, std) over the groups. For example, if reduce by `sha1` over the set of the following `Experiment` objects [`Exp(id=1, sha1=x, f1=0.5)`, `Exp(id=2, sha1=y, f1=0.4)`, `Exp(id=3, sha1=x, f1=0.7)`] and aggregate using `{avg: np.mean, std: np.std}` you will get two `ExperimentAggregate` objects with `{ExpAgg(sha1=x, {metric:f1, avg: 0.6, std=0.01}), ExpAgg(sha1=y, {metric:f1, avg: 0.4, std=0.0})}`. This is supposed to be used for commands like `results`. For this command, previously you could reduce only over `sha1`, now you can reduce over pretty much anything.        

4.  The server talks directly to the backends, you specify the backend while booting it. The backend methods return either Experiment, List[Experiment], or List[ExperimentAggregate] objects. 

5.  We provide some more functionalities now: http://ip:host/v2/classify/id returns an experiment object with the id. Also, you can dump and restore databases using [this utility](https://github.com/sagnik/baseline/blob/feature/xpctl-server/python/xpctl/utils/dump_restore.py)

5.  `cli` makes calls to swagger client, get the objects back, do the necessary things on them and then convert them to dataframes for printing.

This is WIP because the following needs to be done:

- [x]  Update the tests.
- [x]  Check if any dependencies break (xpctl-reporting.py).
- [x]  Add a docker file to run the server.


python/baseline/utils.py
```diff
@@ -19,6 +19,7 @@
 import numpy as np
 import addons
 import collections
+import yaml
```
blester125: yaml is an optional dependency in baseline so this import should be inside the yaml function.

python/baseline/utils.py
```diff
@@ -538,6 +551,12 @@ def write_json(content, filepath):
         json.dump(content, f, indent=True)
 
 
+@exporter
+def write_yaml(content, filepath):
```
blester125: In `baseline.utils` there is now the `str_file` decorator that can handle opening and closing files for you. This lets your function transparently operate on either a file_path (str to a file) or a file like object while you as the programmer only have to write the code as if you are receiving an open file like object.

I think this makes code easier to write, easier to reuse, and more amenable to testing so I think we should start using it more.

This function can be rewritten to use this like so

```python
@str_file(file_path="w")
def write_yaml(content, file_path):
    yaml.dump(content, file_path)
```

dpressel:
xpctl 2.0 is going to be moved to a `mead-ml` repository, so now change here.

PR #304, opened 2019-05-17
added iobes encoded atis dataset

blester125:
This adds an IOBES version of the atis dataset.


dpressel:
I have one of these in a custom dataset index and local system (and a driver configuration). Would like to validate that the uploaded one produces same results

PR #321, opened 2019-06-17
Positional Character-Compositional Embeddings/Transformer LM API Example

dpressel:
- This change adds positional character compositional embeddings
- It also adds an API example using pytorch/ignite to pretrain a Transformer with these embeddings

### DIFF
--- a/python/baseline/pytorch/transformer.py
+++ b/python/baseline/pytorch/transformer.py
@@ -3,7 +3,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 import math
 import numpy as np
-from baseline.pytorch.torchy import pytorch_linear, pytorch_activation, LayerNorm
+from baseline.pytorch.torchy import pytorch_linear, pytorch_activation
 from baseline.pytorch.torchy import pytorch_clone_module, sequence_mask
 from collections import namedtuple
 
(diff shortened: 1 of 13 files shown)


python/baseline/pytorch/transformer.py
```diff
@@ -120,54 +120,29 @@ def forward(self, query, key, value, mask=None):
...
-        :return: An output the same size as the input, but transformed
-        """
-        return self.squeeze(self.dropout(self.act(self.expansion(x))))
-
-
 class TransformerEncoder(nn.Module):
     def __init__(self, num_heads, d_model, pdrop, scale=True, activation_type='relu', d_ff=None):
         super(TransformerEncoder, self).__init__()
         self.d_model = d_model
+
```
blester125: I am getting the following error when trying to make a TransformerEncoder with `d_ff=None`

```traceback
>>> from baseline.pytorch.transformer import TransformerEncoder
>>> te = TransformerEncoder(num_heads=4, d_model=512, pdrop=0.1)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    te = TransformerEncoder(num_heads=4, d_model=512, pdrop=0.1)
  File "/home/blester/dev/work/baseline/python/baseline/pytorch/transformer.py", line 129, in __init__
    self.ffn = nn.Sequential(pytorch_linear(d_model, d_ff),
  File "/home/blester/dev/work/baseline/python/baseline/pytorch/torchy.py", line 200, in pytorch_linear
    l = nn.Linear(in_sz, out_sz)
  File "/home/blester/anaconda3/envs/dl/lib/python3.6/site-packages/torch/nn/modules/linear.py", line 51, in __init__
    self.weight = Parameter(torch.Tensor(out_features, in_features))
TypeError: new() received an invalid combination of arguments - got (NoneType, int), but expected one of:
 * (torch.device device)
 * (torch.Storage storage)
 * (Tensor other)
 * (tuple of ints size, torch.device device)
      didn't match because some of the arguments have invalid types: (NoneType, int)
 * (object data, torch.device device)
      didn't match because some of the arguments have invalid types: (NoneType, int)
```

It seems this is missing a line that defaults `d_ff` like the `FFN` class used to

```python
d_ff = d_ff if d_ff is not None else 4 * d_model
```

dpressel: Yep, you're right. Fixed and base mead file running as expected now

PR #326, opened 2019-06-24
chore: timing logs to console and to files in basedir

blester125:
This PR updates the logging in baseline.

It adds a console logging handler to the timing log that defaults to a level higher than what the timing logger logs at (INFO vs DEBUG). This means that logs don't appear on screen unless you set `TIMING_LOG_LEVEL=DEBUG`

It also creates two extra logging file handlers that create reporting and timing logs in the basedir. This involved shuffling when some arguments are used and a bit of logic duplication from `mead.Task.get_basedir` but it now creates two copies of the logs. One in the basedir (which gets included in the zip) and one in the cwd for back compat

### DIFF
--- a/python/mead/utils.py
+++ b/python/mead/utils.py
@@ -17,7 +17,7 @@ logger = logging.getLogger('mead')
 
 
 @exporter
-def configure_logger(logger_config):
+def configure_logger(logger_config, basedir=None):
     """Use the logger file (logging.json) to configure the log, but overwrite the filename to include the PID
 
     There are reporting and timing loggers that are configured, the latter being used for speed testing.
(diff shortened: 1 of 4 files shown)


python/mead/utils.py
```diff
@@ -29,10 +29,24 @@ def configure_logger(logger_config):
     config = read_config_file_or_json(logger_config, 'logger')
     config['handlers']['reporting_file_handler']['filename'] = 'reporting-{}.log'.format(os.getpid())
     config['handlers']['timing_file_handler']['filename'] = 'timing-{}.log'.format(os.getpid())
+    if basedir is not None:
+        # Create a copy of the file handler so we can keep the log in cwd for back compat
```
dpressel: is there a good reason to do this?

blester125: The none check? or the two handlers?

PR #337, opened 2019-07-31
Added MLM pretraining options to API example

dpressel:
- This adds a masked language model (MLM) objective (AKA denoising auto-encoder) option
- This option cannot be used with character-compositional features yet, only BPE/WordPiece1D
- Masking follows BERT "recipe"

### DIFF
--- a/api-examples/pretrain-transformer-lm.py
+++ b/api-examples/pretrain-transformer-lm.py
@@ -661,7 +661,7 @@ def train():
                     shift_logits = logits[:-1]
                     shift_labels = labels[1:]
                     loss = loss_function(shift_logits, shift_labels)
-                avg_loss.update(loss.item())
+                avg_valid_loss.update(loss.item())
 
         valid_token_loss = avg_valid_loss.avg
         valid_token_ppl = math.exp(valid_token_loss)
--- a/api-examples/pretrain-transformer-lm.py
+++ b/api-examples/pretrain-transformer-lm.py
@@ -517,6 +517,7 @@ def train():
     vocabs = preproc_data['vocabs']
     if args.mlm:
         mask_from = vocabs['x']
+        vocab_size = len(mask_from)
         mask_value = mask_from.get("[MASK]", mask_from.get("<MASK>", -1))
         if mask_value == -1:
             logger.error("We could not find a suitable masking token in the vocab")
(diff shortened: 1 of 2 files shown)


api-examples/pretrain-transformer-lm.py
```diff
@@ -579,11 +589,28 @@ def train():
         for i, batch in enumerate(train_loader):
             x, y = batch
             inputs = {'x': x.to(args.device)}
-            labels = y.to(args.device).transpose(0, 1).contiguous()
+            labels = y.to(args.device)
+            if args.mlm:
+                # Replace 15% of tokens
+                masked_indices = torch.bernoulli(torch.full(labels.shape, 0.15)).byte()
```
blester125: In BERT are all the batches totally packed and there fore we don't have to worry about the thing getting masked is in the padding region?

api-examples/pretrain-transformer-lm.py
```diff
@@ -579,11 +589,28 @@ def train():
...
+                # Anything not masked is 0 so no loss
+                labels[~masked_indices] = 0
+                # Of the masked items, mask 80% of them with [MASK]
+                indices_replaced = torch.bernoulli(torch.full(labels.shape, 0.8)).byte() & masked_indices
+                inputs[indices_replaced] = mask_value
+                # Replace 10% of them with random words, rest preserved for auto-encoding
+                indices_random = torch.bernoulli(torch.full(labels.shape, 0.5)).byte() & masked_indices & ~indices_replaced
+                random_words = torch.randint(args.d_model, labels.shape, dtype=torch.long, device=args.device)
```
blester125: From the help message `args.d_model` is `Model dimension (and embedding dsz)`. This seems to be the wrong thing to use for the high values. This seems like it should be the vocab size

blester125: Are there tokens that we shouldn't include in the random words (Possible by setting the low value in the random? Things like `<PAD>` and `<GO>`?

api-examples/pretrain-transformer-lm.py
dpressel: Yeah, this shouldnt be a concern

api-examples/pretrain-transformer-lm.py
dpressel: Yes, this is wrong, thanks

dpressel: Possibly, this is a good point.  Also we know that this BPE-based masking is sub-optimal.  For now would just like to get an initial attempt in there.

PR #338, opened 2019-08-02
adds a check for pyyaml 5 and change call if there

dpressel:
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation


blester125:
Looks good, tested on pyyaml 5.1 and 3.12, both read the file and neither error'ed or gave a warning

PR #353, opened 2019-09-17
Fix/service-vectorizer

wenshuoliu:
This PR fixes the max-length problem for vectorizers like wordpiece and BPE. Changes:
- Add a `reset` method to subclasses of `Vectorizer` that have a `count` method. This function resets `mxlen` to -1, and `max_seen` to 0. 
- In `services.py`, remove the `set_vectorizer_len` function, and replace it with calling `reset` and then `count` method on vectorizers. The `batch_input` function doesn't keep track of `mxlen` any more. 
- Remove the truncation in `batch_input` function, as well as `get_vectorizer_len` function. 
- Add `count` and `reset` to the wordpiece1d vectorizer of the `embed_bert.py`. 

The changes are tested on classifier, tagger, embedding and encoder-decoder services, as well as the new bert addon. The LM part of the `services.py` was not touched.

### DIFF
--- a/api-examples/lm-text.py
+++ b/api-examples/lm-text.py
@@ -1,9 +1,10 @@
 import baseline as bl
 import argparse
 import os
-parser = argparse.ArgumentParser(description='Classify text with a model')
-parser.add_argument('--model', help='A classifier model', required=True, type=str)
-parser.add_argument('--text', help='raw value', type=str)
+parser = argparse.ArgumentParser(description='Generate subsequent text by repeatedly predicting the next word from a '
+                                             'language model')
+parser.add_argument('--model', help='A language model', required=True, type=str)
+parser.add_argument('--text', help='raw value, a string', type=str)
 parser.add_argument('--device', help='device')
 
 
@@ -14,7 +15,7 @@ if os.path.exists(args.text) and os.path.isfile(args.text):
     with open(args.text, 'r') as f:
         for line in f:
             text = line.strip().split()
-            texts += [text]
+            texts += text  # consider the whole file as a long sequence input
 
 else:
     texts = args.text.split()
--- a/api-examples/lm-text.py
+++ b/api-examples/lm-text.py
@@ -6,6 +6,7 @@ parser = argparse.ArgumentParser(description='Generate subsequent text by repeat
 parser.add_argument('--model', help='A language model', required=True, type=str)
 parser.add_argument('--text', help='raw value, a string', type=str)
 parser.add_argument('--device', help='device')
+parser.add_argument('--mxlen', type=int, help='how many tokens will be generated', default=10)
 
 
 args = parser.parse_known_args()[0]
@@ -15,7 +16,7 @@ if os.path.exists(args.text) and os.path.isfile(args.text):
     with open(args.text, 'r') as f:
         for line in f:
             text = line.strip().split()
-            texts += text  # consider the whole file as a long sequence input
+            texts += [text]
 
 else:
     texts = args.text.split()
@@ -23,4 +24,4 @@ else:
 print(texts)
 
 m = bl.LanguageModelService.load(args.model, device=args.device)
-print(m.predict(texts))
+print(m.predict(texts, mxlen=args.mxlen))
--- a/python/addons/custom_vectorizers.py
+++ b/python/addons/custom_vectorizers.py
@@ -1,5 +1,6 @@
 import numpy as np
 from collections import Counter
+import tempfile
 from baseline.utils import Offsets
 from baseline.vectorizers import register_vectorizer, AbstractVectorizer
 
@@ -82,7 +83,6 @@ class SavableFastBPE(object):
 
     def __setstate__(self, state):
         from fastBPE import fastBPE
-        import tempfile
         with tempfile.NamedTemporaryFile() as codes, tempfile.NamedTemporaryFile() as vocab:
             codes.write(state['codes'])
             vocab.write(state['vocab'])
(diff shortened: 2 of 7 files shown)


python/addons/embed_bert.py
```diff
@@ -1249,6 +1249,19 @@ def __init__(self, **kwargs):
         self.wordpiece_tok = WordpieceTokenizer(load_vocab(kwargs.get('vocab_file')))
         self.mxlen = kwargs.get('mxlen', -1)
 
+    def count(self, tokens):
```
blester125: Does this mean that the `count` method should be moved into the `AbstractVectorizer` instead of being in the `Token1DVectorizer`?

python/baseline/services.py
```diff
@@ -238,8 +195,11 @@ def predict(self, tokens, preproc=None):
         """
         if preproc is not None:
             logger.warning("Warning: Passing `preproc` to `ClassifierService.predict` is deprecated.")
-        tokens_seq, mxlen, mxwlen = self.batch_input(tokens)
-        self.set_vectorizer_lens(mxlen, mxwlen)
+        tokens_seq = self.batch_input(tokens)
```
blester125: I know this is the name we used before but this would be clearer if the name was something like `tokens_batch` given that it is returned form the batch_input function

python/baseline/services.py
```diff
@@ -402,8 +331,11 @@ def predict(self, tokens, **kwargs):
         if not export_mapping:
             export_mapping = {'tokens': 'text'}
         label_field = kwargs.get('label', 'label')
-        tokens_seq, mxlen, mxwlen = self.batch_input(tokens)
-        self.set_vectorizer_lens(mxlen, mxwlen)
+        tokens_seq = self.batch_input(tokens)
+        for _, vectorizer in self.vectorizers.items():
```
blester125: This should probably get moved into a method given that this loop is done in so many places.

python/baseline/vectorizers.py
```diff
@@ -80,6 +80,10 @@ def count(self, tokens):
         self.max_seen = max(self.max_seen, seen)
         return counter
 
+    def reset(self):
```
blester125: Like mentioned above when adding the count method to the bert vectorizer it seems like it would make sense to add these methods to the `AbstractVectorizer`

wenshuoliu:
Updates: 
- adressed comments from @blester125 
- updated lm part of the service and tested with wordpiece1d vectorizer
- possibly fixed a bug in `SeqWordCharDataFeed`

The vectorizer based on fastBPE was added to addon `custom_vectorizers.py`. But when testing on training an lm, I got the error "can't pickle a c object" when program tries to save the vectorizer. Will come back and look into it.

api-examples/lm-text.py
```diff
@@ -14,7 +15,7 @@
     with open(args.text, 'r') as f:
         for line in f:
             text = line.strip().split()
-            texts += [text]
+            texts += text  # consider the whole file as a long sequence input
```
blester125: I don't think that this is what we want to do. It seems like each line would be a sentence in the batch that we would predict the next token of?

python/baseline/services.py
```diff
@@ -58,62 +58,31 @@ def batch_input(self, tokens):
...
-                if hasattr(vectorizer, 'mxlen'):
-                    mxlen = vectorizer.mxlen
-                if hasattr(vectorizer, 'mxwlen'):
-                    mxwlen = vectorizer.mxwlen
-        return mxlen, mxwlen
+        tokens_batch = (tokens,) if isinstance(tokens[0], six.string_types) else tokens
+        return tokens_batch
 
-    def vectorize(self, tokens_seq):
+    def prepare_vectorizers(self, tokens):
```
blester125: I think this method is doing to much. It is batching the input and then preparing the vectorizers. This method should probably take in the `token_batch` instead of creating it. so more like

```python
def prepare_vectorizers(self, token_batch):
     for vectorizer in self.vectorizers.values():
            vectorizer.reset()
            for tokens in tokens_batch:
                _ = vectorizer.count(tokens)
```

Then use it later with

```python
token_batch = self.batch_input(tokens)
self.prepare_vectorizers(self, token_batch)
```

python/baseline/services.py
```diff
@@ -454,18 +386,17 @@ def predict(self, tokens, **kwargs):
         mxwlen = kwargs.get('mxwlen', 40)
 
         for k, vectorizer in self.vectorizers.items():
```
blester125: Why is the `vectorizer.mxlen` still getting manually set here? Shouldn't this be handled by the `prepare_vectorizers` method?

api-examples/lm-text.py
wenshuoliu: In that way, I would add `batch_input` in the `LanguageModelService`, right?

python/baseline/services.py
wenshuoliu: agreed

python/addons/custom_vectorizers.py
```diff
@@ -81,6 +81,8 @@ def __getstate__(self):
         return {'codes': self.codes, 'vocab': self.vocab}
 
     def __setstate__(self, state):
+        from fastBPE import fastBPE
+        import tempfile
```
blester125: tempfile is a std library import so it should go at the top of the file we only do imports like this for optional deps like fastBPE

wenshuoliu:
Update: 
- Wrap fastBPE tokenizer so it can be pickled. 
- Batch the input of lm service.

api-examples/lm-text.py
wenshuoliu: I changed it to a batch-input way.

python/baseline/services.py
wenshuoliu: made the change that `mxlen` is set by `prepare_vectorizers`

PR #358, opened 2019-10-12
WIP: A bert ner implementation in pytorch

blester125:
This is a bert model that can do NER as they do in the paper where there is a linear layer on top of the model and the label is read from the starting token.

This is accomplished via a few custom vectorizers. There is a new label vectorizer that when it outputs labels it outputs the true label of the first token in a subword string and `PAD` for the rest of them. There is also a custom vectorizers that outputs a list of indices that point to the first token in each word. The normal vectorizer produces the tokens that bert expects to see. The index vector pointing at the start words is used in a `torch.gather` to select the starting tokens (or their associated vector from the unaries) from the other tenors. This results in a `[B, T, H]` matrix where the tokens are only in T if they are a staring token.

Doing this gather (instead of just relaying on the masking in the tags in conjunction with the `padding_index` in the loss to only count the heads) give us a nice advantage that the outputs can be feed into things like the constrained decoder and they just work. It also means that the labels output are aligned with the real tokens (before wordpiece breaks them up) so we don't need to do any sort of detokeniztion in the service to label the words.

This model works when it is reloaded as a service but I had to make most of the change from [this PR](https://github.com/dpressel/mead-baseline/pull/353) to get it to work.

There are also a few bug fixes included that were needed to make it run

 * The TaggerGreedyDecoder was doing the softmax on the wrong dim making the mask all zero
 * The mask above was getting unsqueezed twice so it was the wrong shape
 * The TaggerGreedyDecoder wasn't using the SequenceCriterion that handles folding the B, T correctly
 * The SequenceCriterion in the TaggerGreedyDecoder was hard coded to do batch loss but in bert they mostly do token loss so I had to expose that
 * The way the preds were created in the TaggerGreedyDocder were generated ragged batches so I updated that to create a [B, T] one.
 * The Bert Embedding that returned sequences was hard coded to detach the vectors (no fine tuning) I make this changeable via the config file
 * There was a bug in Bert Embedding that returned Sequences where the multiple layers would be concatenated and means on the last dim if the operator was not `concat`. This results in tensor with the shape `[B, T, 1]` which isn't what you want. I updated it so if we are taking the mean of the vector across the selected layers we do an unsqeeze before the concat, this creates a vector of shape `[B, T, H, #Layers]` then a mean on the last dim (the layers) gives us `[B, T, H]` which is what we want. 

There were a few tiny tweaks I had to make to baseline proper for this to work.

 * When evaluation the tagger I had to use the `y_lengths` instead of the lengths to know where to cut the batch. This is because the lengths key is about the size of the input but the eval needs to know the size of the output. This is normally the same but in a case where the vectorizer can expand the number of tokens in the input it is not.I left a backoff to the `lengths` for backwards compat
 * I needed a way to get the start of token indices (called head indices in the code) into the batch_dict but I didin't want them to be embeddinged (even if I made a passthrough embedding they would still get concatenated on the bert embeddings which I don't want) so I made a few tweaks so that is a vectorizer says its embeddings is `None` then its output just gets added to the batch_dict but not the embeddings dict. If we don't like this change I explored having the vectorizer output both the token and if it was the start or not and then in the `make_input` it would extract this into two different tensors. This can be seen in the `bert2` model. This way was very slow howeverm a single epoch took about 950 seconds compared to 170 the `bert` tagger with multiple vectorizers takes

I haven't done a large parameter sweep yet but just using some hps in the range they mention in the paper (using `bert-base-uncased`) I was seeing dev score in the 95, this isn't at the level of their paper but I think with tweaks on the HPs and using `bert-large` (and maybe the cased version) we should see better results.

One question I haven't decided on yet is what to do about the `do_basic_tokenization` parameter that the `pytorch_pretrained_bert` tokenizer has. Currently that param is one and it results in things like `3/1/97` which is a token in conll become `['3', '/', '1', '/', '97']`. The output is still aligned because the vectorizers treat this like the token was split by wordpiece but it might make more sense to turn this parameter off and just like the wordpiece handle it?

Implementing this model and tracking all the indices and what not is rather complex, if someone has a simpler solution I would love to discuss it.

### DIFF
--- a/python/addons/embed_bert_pytorch.py
+++ b/python/addons/embed_bert_pytorch.py
@@ -98,121 +98,9 @@ class DictWordPieceVectoizer1D(WordPieceVectorizer1D):
         return super().iterable(_token_iterator(self, tokens))
 
 
-def mask_to_index(mask):
-    """Convert a 1d mask to a 1d list of indices of the 1's. the result is the same shape as the mask."""
-    idx = mask.nonzero()[0]
-    return np.concatenate((idx, np.zeros((len(mask) - len(idx),), dtype=mask.dtype)))
-
-
-@register_vectorizer(name='wordpiece1d-with-heads')
-class WordPieceVectorizer1DWithHeads(WordPieceVectorizer1D):
-    """This vectorizer produces the tokens and the index of the heads at the same time.
-
-    This outputs a matrix of size [2, T] where [0, T] is the tokens and [1, T] is the mask
-    """
-
-    def iterable(self, tokens):
-        yield '[CLS]', 0
-        for tok in tokens:
-            if tok == '<unk>':
-                yield '[UNK]', 1
-            elif tok == '<EOS>':
-                yield '[SEP]', 0
-            else:
-                for i, subtok in enumerate(self.tokenizer.tokenize(tok)):
-                    if i == 0:
-                        yield subtok, 1
-                    else:
-                        yield subtok, 0
-        yield '[SEP]', 0
-
-    def _next_element(self, tokens, vocab):
-        for atom, head in self.iterable(tokens):
-            value = vocab.get(atom, vocab['[UNK]'])
-            yield value, head
-
-    def run(self, tokens, vocab):
-        self.mxlen = self.max_seen if self.mxlen < 0 else self.mxlen
-        vec1d = np.zeros(self.mxlen, dtype=np.long)
-        heads = np.zeros(self.mxlen, dtype=np.long)
-        for i, (atom, head) in enumerate(self._next_element(tokens, vocab)):
-            if i == self.mxlen:
-                i -= 1
-                break
-            vec1d[i] = atom
-            heads[i] = head
-        heads = mask_to_index(heads)
-        valid_length = i + 1
-        return np.stack((vec1d, heads)), valid_length
-
-
-@register_vectorizer(name="dict-wordpiece1d-with-heads")
-class DictWordPieceVectoizer1DWithHeads(WordPieceVectorizer1DWithHeads):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self.fields = listify(kwargs.get('fields', 'text'))
-        self.delim = kwargs.get('token_delim', '~~')
-
-    def iterable(self, tokens):
-        return super().iterable(_token_iterator(self, tokens))
-
-
-@register_vectorizer(name='wordpiece1d-with-heads-dict')
-class WordPieceVectorizer1DWithHeadsDict(WordPieceVectorizer1D):
-    """This vectorizer produces the tokens and the index of the heads at the same time.
-
-    In this one the results are output as a dict
-    """
-
-    def iterable(self, tokens):
-        yield '[CLS]', 0
-        for tok in tokens:
-            if tok == '<unk>':
-                yield '[UNK]', 1
-            elif tok == '<EOS>':
-                yield '[SEP]', 0
-            else:
-                for i, subtok in enumerate(self.tokenizer.tokenize(tok)):
-                    if i == 0:
-                        yield subtok, 1
-                    else:
-                        yield subtok, 0
-        yield '[SEP]', 0
-
-    def _next_element(self, tokens, vocab):
-        for atom, head in self.iterable(tokens):
-            value = vocab.get(atom, vocab['[UNK]'])
-            yield value, head
-
-    def run(self, tokens, vocab):
-        self.mxlen = self.max_seen if self.mxlen < 0 else self.mxlen
-        vec1d = np.zeros(self.mxlen, dtype=np.long)
-        heads = np.zeros(self.mxlen, dtype=np.long)
-        for i, (atom, head) in enumerate(self._next_element(tokens, vocab)):
-            if i == self.mxlen:
-                i -= 1
-                break
-            vec1d[i] = atom
-            heads[i] = head
-        heads = mask_to_index(heads)
-        valid_length = i + 1
-        return {'tokens': vec1d, 'heads': heads}, valid_length
-
-
-@register_vectorizer(name="dict-wordpiece1d-with-heads-dict")
-class DictWordPieceVectoizer1DWithHeadsDict(WordPieceVectorizer1DWithHeadsDict):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self.fields = listify(kwargs.get('fields', 'text'))
-        self.delim = kwargs.get('token_delim', '~~')
-
-    def iterable(self, tokens):
-        return super().iterable(_token_iterator(self, tokens))
-
-
-@register_vectorizer(name='heads-wordpiece1d')
-class HeadedWordPieceVectorizer1D(WordPieceVectorizer1D):
-    """Output a list of indices that are the heads of the subword tokens."""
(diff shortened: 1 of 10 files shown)


blester125:
One odd artifact of this implemention is the vectorizer that produces the head indices has a count that counts the actual tokens not the heads, i was thinking that this is needed because if we just count heads then the max seen will be too low compared the the normal wordpiece vectorizer

One way to fix this would be to have the vectorizer run produce a mask and convert it to indices like the joint text and head vectorizer does. This won't require a count that counts something different than what the run produces.

I actually haven't tried it with having this custom count, when i get back I'll try it where the count is based on the heads and see it that actually does break it, now I'm thinking it might not and this comment may be moot.

In general I originally developed this where the mxlens on the vectorizers were set by the hard coded mxlen, now that they have counts there might be some hacks I can remove, I'll investigate these soon

python/addons/embed_bert_pytorch.py
```diff
@@ -70,6 +82,263 @@ def run(self, tokens, vocab):
...
+                        yield j  # Head
+                    else:
+                        j += 1
+            j += 1
+
+    def _next_element(self, tokens, vocab):
+        for atom in self.iterable(tokens):
+            yield atom
+
+    def run(self, tokens, vocab):
```
blester125: This needs to be test for the case where mxlen is less than the length of the example. 

It might need to refactor so instead of the i == mxlen break check it might need to check if the head index is >= mxlen. Or else it might need to do a clamp so one of the indices isn't larger than the mxlen

python/addons/embed_bert_pytorch.py
```diff
@@ -132,3 +413,218 @@ def __init__(self, name, **kwargs):
...
+        if self.gpu:
+            lengths = lengths.cuda()
+        example_dict['lengths'] = lengths
+        # The vectorizer outputs a [2, T] tensor which is stacked into [B, 2, T].
+        # The first index in the 2 dim is the tokens and the second is the heads
+        for key in self.embeddings.keys():
+            example_dict[key] = self.input_tensor(key, batch_dict, perm_idx)
+
+        # This assumes all heads are aligned
+        heads = [k for k in batch_dict.keys() if k.endswith('_heads')][0]
```
blester125: Doing thia search everytime is waste full. Can the name if the key be cached or passed in?

python/addons/embed_bert_pytorch.py
```diff
@@ -132,3 +413,218 @@ def __init__(self, name, **kwargs):
...
+    def make_input(self, batch_dict):
+
+        example_dict = dict({})
+        lengths = torch.from_numpy(batch_dict[self.lengths_key])
+        lengths, perm_idx = lengths.sort(0, descending=True)
+
+        if self.gpu:
+            lengths = lengths.cuda()
+        example_dict['lengths'] = lengths
+        # The vectorizer outputs a [2, T] tensor which is stacked into [B, 2, T].
```
blester125: This comment doesn't apply to this make input, its a copy paste error from a make input below

python/baseline/reader.py
```diff
@@ -335,7 +335,7 @@ def load(self, filename, vocabs, batchsz, shuffle=False, sort_key=None):
         for i, example_tokens in enumerate(texts):
             example = {}
             for k, vectorizer in self.vectorizers.items():
-                example[k], lengths = vectorizer.run(example_tokens, vocabs[k])
+                example[k], lengths = vectorizer.run(example_tokens, vocabs.get(k))
```
blester125: This is a change to facilitate the idea that a feature might just be passed to the example dict instead of being passed through an embeddeding

python/eight_mile/pytorch/layers.py
```diff
@@ -831,18 +831,55 @@ def viterbi(unary, trans, lengths, start_idx, end_idx, norm = lambda x, y: x):
...
 
-    def __init__(self, num_tags, constraint_mask=None, batch_first=True):
+    def __init__(self, num_tags, constraint_mask=None, batch_first=True, reduction='batch'):
         super(TaggerGreedyDecoder, self).__init__()
         self.num_tags = num_tags
 
         if constraint_mask is not None:
-            self.constraint_mask = F.log_softmax(torch.zeros(constraint_mask.shape).masked_fill(constraint_mask, -1e4), dim=0)
-            self.register_buffer('constraint_mask', self.constraint_mask.unsqueeze(0))
+            constraint_mask = F.log_softmax(torch.zeros(constraint_mask.shape).masked_fill(constraint_mask, -1e4), dim=1)
```
blester125: Should add to the doc string that the constraint_mask should have shape `[1, N, N]`

python/addons/embed_bert_pytorch.py
```diff
@@ -70,6 +82,263 @@ def run(self, tokens, vocab):
...
+    def count(self, tokens):
+        return super().count(_token_iterator(self, tokens))
+
+
+class LabelHeadedWordPieceVectorizer1D(DictHeadedWordPieceVectorizer1D):
+    """Convert labels so just the heads have labels, other labels are PAD"""
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.fields = listify(kwargs.get('fields', ['text', 'y']))
+        self.delim = kwargs.get('token_delim', '~~')
```
blester125: AFAIK this is one of the first vectorizers that actually use multiple fields from the conll file at the same time. I had to change the token delim from `@@` because there was a `@` as a token. This resulted in token that looked like `@@@O` and a split of `[''. '@O']` the empty token caused problems because the token seqs for this vectorizer wasn't aligned with the others

python/addons/embed_bert_pytorch.py
```diff
@@ -70,6 +82,263 @@ def run(self, tokens, vocab):
...
+                break
+            vec1d[i] = atom
+            heads[i] = head
+        heads = mask_to_index(heads)
+        valid_length = i + 1
+        return np.stack((vec1d, heads)), valid_length
+
+
+@register_vectorizer(name="dict-wordpiece1d-with-heads")
+class DictWordPieceVectoizer1DWithHeads(WordPieceVectorizer1DWithHeads):
```
dpressel: typo

python/addons/embed_bert_pytorch.py
```diff
@@ -70,6 +82,263 @@ def run(self, tokens, vocab):
...
+                break
+            vec1d[i] = atom
+            heads[i] = head
+        heads = mask_to_index(heads)
+        valid_length = i + 1
+        return {'tokens': vec1d, 'heads': heads}, valid_length
+
+
+@register_vectorizer(name="dict-wordpiece1d-with-heads-dict")
+class DictWordPieceVectoizer1DWithHeadsDict(WordPieceVectorizer1DWithHeadsDict):
```
dpressel: typo

blester125:
Another question to address is if all this head selection code should be pushed into an embeddeding model instead of being part of the actual tagger model.

The way it is right now (in the tagger model) allows flexibility in the decoder model but the transducer model is stuck as a pass through.

If the head selection was pushed into the embeddeding layer it would allow custom transducer models on top and would allow easy replication of the feature based NER experiment in the paper. As it is currently this model can only do the fine-tuning NER tagging from the paper.

A complication of pushing it into the embeddeding is how to get both of the inputs (the token indices and the head indices) into the embeddings. Right now its very one vectorizer on embedding

dpressel:
Superceded by #464

PR #361, opened 2019-10-15
TF 2.0 works in eager mode

dpressel:


python/baseline/tf/tagger/model.py
```diff
@@ -163,6 +154,17 @@ def make_input(self, batch_dict, train=False):
             feed_dict[self.y] = y
         return feed_dict
 
+    def __call__(self, *args, **kwargs):
```
blester125: Somethings seem to have a `call` and others `__call__` is that a `tf.keras.Model` vs `tf.keras.layers.Layer` thing?

PR #366, opened 2019-10-16
Chore/positional mixin

blester125:
This PR updates the embeddings, there are now the idea of embeddings as layers which live in eight_mile, these are tensor in tensor outs layers. There are also model wrappers, these live in baseline and are the ones that are registered.

There is also a positional mixin. This can be used to add positional embeddings (learned or sin based ones) to other embeddings. Unfortunately I also hard to create a positional layer that can kind of stand on it's own so that it can be used in the embeddings that are `tf.keras.Model` instead of `tf.keras.layers.Later`

I also removed the Large LUT embeddings for now because all the training functions have lost the init steps and the embeddings requires a placeholder.

### DIFF
--- a/python/baseline/tf/embeddings.py
+++ b/python/baseline/tf/embeddings.py
@@ -1,97 +1,222 @@
+import logging
+from eight_mile.embeddings import register_embeddings
 from eight_mile.tf.embeddings import *
 import tensorflow as tf
 
 
-def stacked_lstm(hsz, pdrop, nlayers, variational=False, training=False):
-    """Produce a stack of LSTMs with dropout performed on all but the last layer.
+logger = logging.getLogger('baseline')
 
-    :param hsz: (``int``) The number of hidden units per LSTM
-    :param pdrop: (``int``) The probability of dropping a unit value during dropout
-    :param nlayers: (``int``) The number of layers of LSTMs to stack
-    :param variational (``bool``) variational recurrence is on
-    :param training (``bool``) Are we training? (defaults to ``False``)
-    :return: a stacked cell
+
+class TensorFlowEmbeddingsBaselineModel(tf.keras.Model):
+    """This provides a base for TensorFlow embeddings sub-graphs that includes the placeholders
+
     """
-    if variational:
-        return tf.contrib.rnn.MultiRNNCell(
-            [lstm_cell_w_dropout(hsz, pdrop, variational=variational, training=training) for _ in range(nlayers)],
-            state_is_tuple=True
-        )
-    return tf.contrib.rnn.MultiRNNCell(
-        [lstm_cell_w_dropout(hsz, pdrop, training=training) if i < nlayers - 1 else lstm_cell(hsz) for i in range(nlayers)],
-        state_is_tuple=True
-    )
+    def __init__(self, name=None, **kwargs):
+        """Constructor
+        """
+        super().__init__()
+        self._record_state(**kwargs)
+        self._name = name
+        self.embedding_layer = None
+
+    def detached_ref(self):
+        """This will detach any attached input and reference the same sub-graph otherwise
+
+        :return:
+        """
+        if getattr(self.embedding_layer, '_weights', None) is not None:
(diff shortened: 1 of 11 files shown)


python/baseline/tf/embeddings.py
```diff
@@ -1,75 +1,216 @@
...
+        """
+        return self.embedding_layer.get_vsz()
+
+    def encode(self, x):
+        """This instantiates the sub-graph for this object and returns the output node
+
+        :return:
+        """
+        if x is None:
+            x = self.create_placehodler(self._name)
```
dpressel: typo

python/eight_mile/pytorch/layers.py
```diff
@@ -137,7 +137,7 @@ def rnn_bi_hidden(output, output_state):
         output_state = tuple(_cat_dir(h) for h in output_state)
     else:
         output_state = _cat_dir(output_state)
-    return output, output_state
+    return output_state
```
dpressel: this is going away

python/baseline/tf/embeddings.py
```diff
@@ -1,75 +1,216 @@
...
+
+    def call(self, x):
+        return self.encode(x)
+
+    def get_feed_dict(self):
+        """Return a feed dict that is needed to initialize this embeddings."""
+        return self.embedding_layer.get_feed_dict()
+
+    @classmethod
+    def create_placehodler(cls, name):
```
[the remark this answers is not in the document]
blester125: fixed

python/baseline/tf/embeddings.py
blester125: fixed

python/eight_mile/pytorch/layers.py
blester125: What should we do about that? Should this wait and be updated once you 2.0 branch is merged in? Or should I make test and make this PR against that branch?

blester125:
This needs to be updated once https://github.com/dpressel/mead-baseline/pull/361 is merged in

blester125:
This PR does the following

 * Updates the embeddings in `eight_mile/(tf|pytorch)/embeddings.py` so that they are layers that are tensor in, tensor out
 * Updates the embeddings in `baseline/(tf|pytorch)/embeddings.py` so that they are Models that get registered (and in tf's case it handles place holder creation)
 * Updates the TF2 BiLSTMEncoder implementations to return the the same things as the TF1 and Pytorch implementations (cating the forward and backward hidden state)
 * Updates to the the Pytorch CharConv embeddings so it matches tf
 * Updates to the TF CharLSTM to handle the fact that tf2 uses cudnn under the hood
 * General updates to allow for tf1 and tf2 compatibility
 * Move the code that calcuates the output size of CharConv into a shared area so both can use it (added tests too)
 * Creates Mixins for the positional and learned positional embeddings

python/eight_mile/tf/embeddings.py
```diff
@@ -143,234 +69,174 @@ def __init__(self, name, **kwargs):
...
+        B = shape[0]
+        T = shape[1]
+        W = shape[2]
+        flat_chars = tf.reshape(x, [-1, W])
+        embed_chars = self.embed(flat_chars)
+
+        # Calculate the lengths of each word
+        word_lengths = tf.reduce_sum(tf.cast(tf.equal(flat_chars, Offsets.PAD), tf.int32), axis=1)
+
+        # cuDNN throws an error if there is an input with a length of 0, this happens when the "word"
```
blester125: A design question is if we should push this code about patching lengths of 0 into the LSTM encoder itself. When a user is using the LSTMEncoder layer I would imagine it is rare that they have a thing in the batch that is length 0 so it might be a waste of time to handle this for all LSTMEncoder calls but on the other hand this is an esoteric error so handling it under the hood would save users a lot of time if they ever hit that issue

blester125: From discussion we decided to leave this here with comments for now and eventually push it in if other people report case where they get the error

PR #367, opened 2019-10-16
fix: pytorch classification models

blester125:
Outside of the pytorch ConvModel all the classification models still used the old format where the `init_pool` method returned a size and the `pool` method did the actual pooling instead of `init_pool` returning a pooling model with a `output_dim` attribute that does the pooling in it's forward call.

This PR fixes the classification models so that they all work again.


blester125:
This needs to be updated once https://github.com/dpressel/mead-baseline/pull/361 is merged

blester125:
This PR does the following:

 * Updates baseline pytorch classification models to work with the new format where `init_pool` returns a `nn.Module` instead of a `int` (output size)
 * Updates the eight mile pytorch layers:
   * Updates the the LSTM classes to match the tf ones
   * Adds a max and mean pooling layer
   * Fixes some bugs in the DenseStack module
   * Adds a composite module that runs multiple models and concats the values
 * Updates the tf NBow base class so that the it actually does force a stacked layer for nbow models

PR #369, opened 2019-10-17
fix: get the other tf classify models running

blester125:
This PR updates the tf classify models to fix the nbowmax and composite models.

I was able to get them to train with both tf2 and tf1

I also tested and it is working where the pool model is a mix of models that need lengths and those that don't

### DIFF
--- a/api-examples/eager-tagger.py
+++ b/api-examples/eager-tagger.py
@@ -1,4 +1,5 @@
 
+import os
 from eight_mile.utils import listify
 import baseline
 import eight_mile.tf.layers as L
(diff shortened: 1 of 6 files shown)


api-examples/eager-tagger.py
```diff
@@ -55,11 +56,11 @@ def to_tensors(ts):
     return np.stack(X), np.stack(Xch), np.stack(y)
 
 
-VSM_MODEL = '/home/dpressel/.bl-data/dce69c404025a8312c323197347695e81fd529fc/glove.twitter.27B.200d.txt'
+VSM_MODEL = os.path.expanduser('~/.bl-data/dce69c404025a8312c323197347695e81fd529fc/glove.twitter.27B.200d.txt')
 
-TS = '/home/dpressel/dev/work/baseline/data/oct27.train'
-VS = '/home/dpressel/dev/work/baseline/data/oct27.dev'
-ES = '/home/dpressel/dev/work/baseline/data/oct27.test'
+TS = os.path.expanduser('~/dev/work/baseline/data/oct27.train')
```
dpressel: this should be changed anyway to use argparse

PR #376, opened 2019-10-23
WIP: Feature/layer transformer

dpressel:
- Creates and identical layer in PyTorch and TF between Transformer primitives
- Adds a standalone eager mode training program for TLMs in TF that mimics pytorch one


python/eight_mile/tf/layers.py
```diff
@@ -938,27 +919,27 @@ def requires_length(self):
         return False
 
 
-def scaled_dot_product_attention(query, key, value, pdrop=0.0, mask=None):
+def scaled_dot_product_attention(query, key, value, mask=None, pdrop=0.0):
```
blester125: It looks like these attentions aren't doing dropout like the pytorch ones do, this seems like a large mismatch, similarly there is an api mismatch where pytorch takes a dropout object and this is taking a float

python/eight_mile/tf/layers.py
```diff
@@ -1347,27 +1324,40 @@ def neg_log_loss(self, unary, tags, lengths):
...
-        super(LangSequenceModel, self).__init__(name=name)
+        super().__init__(name=name)
         if isinstance(embeddings, dict):
             self.embed_model = EmbeddingsStack(embeddings)
         else:
             assert isinstance(embeddings, EmbeddingsStack)
             self.embed_model = embeddings
         self.transducer_model = transducer
-
+        if hasattr(transducer, 'requires_state') and transducer.requires_state:
```
blester125: `if getattr(transducer, 'requires_state', False):` also works, is this more explicit?

python/baseline/pytorch/seq2seq/decoders.py
```diff
@@ -115,14 +106,14 @@ def __init__(self, tgt_embeddings, **kwargs):
...
-        self.decoder_rnn = pytorch_rnn_cell(dsz, self.hsz, rnntype, layers, pdrop)
+        self.decoder_rnn = rnn_cell(dsz, self.hsz, rnntype, layers, pdrop)
         self.dropout = torch.nn.Dropout(pdrop)
         self.init_attn(**kwargs)
 
         do_weight_tying = bool(kwargs.get('tie_weights', False))
         is_valid_tying = self.hsz == self.tgt_embeddings.get_dsz()
 
-        self.preds = pytorch_linear(self.hsz, self.tgt_embeddings.get_vsz())
+        self.preds = torch.nn.Linear(self.hsz, self.tgt_embeddings.get_vsz())
```
blester125: Using `torch.nn.Linear` directly loses the initialization we do in `pytorch_linear` like setting the bias to `0`. Do we want to remove that?

python/eight_mile/pytorch/layers.py
```diff
@@ -1434,3 +1516,227 @@ def call(self, inputs):
...
+            for i in range(mxlen - 1):
+                probs, extra = self.step(paths, extra)
+                V = probs.shape[-1]
+                probs = probs.view((bsz, self.K, V))  # [B, K, V]
+                if i > 0:
+                    # This mask is for all beams that are done.
+                    done_mask = (lengths != 0).unsqueeze(-1)  # [B, K, 1]
+                    # Can creating this mask be moved out of the loop? It never changes but we don't have V
+                    # This mask selects the EOS token
+                    eos_mask = torch.zeros((1, 1, V), dtype=MASK_TYPE, device=device)
```
blester125: I thought the plan was to remove the `MASK_TYPE` and have the dtype based on the dtype for `done_mask` which should correctly handle the type for different versions of pytorch? Or did that not work?

python/eight_mile/tf/layers.py
```diff
@@ -74,6 +74,10 @@ def swish(x):
     return x*tf.nn.sigmoid(x)
 
 
+def masked_fill(t, mask, value):
```
blester125: This is smart, we should have done this a long time ago lol

python/eight_mile/tf/layers.py
```diff
@@ -1544,3 +1575,404 @@ def with_device(*args, **kwargs):
             return func(*args, **kwargs)
         return func(*args, **kwargs)
     return with_device
+
+
+class BaseAttention(tf.keras.layers.Layer):
```
blester125: Have there been any papers on if using the kqv form of attention works better for this sort of vector to sequence attention used in seq2seq?

python/eight_mile/pytorch/layers.py
dpressel: yep. its on the list!  moving slower than I hoped

blester125:
I have updated this PR to get seq2seq working for tf1 and tf2. There were a bunch of weird quirks to get it to work.

I had to remove the weight tying for now, I am working on getting it back working but I am going to merge this for now

PR #393, opened 2019-11-13
Blester/fix/no lengths for preproc

sagnik:
This change makes it possible for utterance lengths not be sent. I have tested by running `./test_exporters.sh` (which was modified to be compatible for changes in `mead-export`)


sagnik:
@blester125 and I had some discussions about this. I am merging this.

PR #398, opened 2019-11-18
WIP: fix: mxlen is now totally calculated server side

blester125:
This PR updates models that are processed server side so that they have no hard mxlen cap, they will handle any size you throw at them.


sagnik:
This MR 1. makes the remote tf serving models disregard the mxlen set by the trainer 2. doesn't handle the `mxwlen` in a way similar to `mxlen`. This does not create a problem for now because the chars are used in conv and not an RNN, but will break things when used in an RNN. I am merging this: the second problem can be solved later, but the first one is a policy decision.

PR #404, opened 2019-12-05
fix: use correct output delimiter when input delim is none in conll converter

blester125:
The bug that [this PR](https://github.com/dpressel/mead-baseline/pull/403) was fixing was that when the delimiter passed to the conll converter was `None` then when writing the output file it try to join on a None which would fail. The fix above was to pass a delimiter that defaults to space.

This makes the script more annoying to use because it no longer handles both space separated and tab separated conll files transparently. Defaulting the delim to `' '` causes `\t` delimited files to fail. This change leaves the ability to pass a delimiter via command line (it isn't clear how you would pass a tab though) but it defaults it to `None`. It fixes the original bug in the actual converter code itself, if there is no delimiter (which is fine for reading the conll file) it defaults to using a space for the output delimiter. This fixes the same bug without breaking the ability to handle tab separated conll files. 

Also the changes in that PR we removing the last 4 characters from the input file name before adding the suffix (`.iobes`) to the file. This seems to be for removing the `.bio` at the end of a file. This seems overly zealous and is you pass in a file like `example.conll` you get an output file of `example.c.iobes`. This PR updates it so it only does this striping if the file actually ends with `.bio`

### DIFF
--- a/api-examples/bio_to_iobes.py
+++ b/api-examples/bio_to_iobes.py
@@ -10,24 +10,28 @@ parser.add_argument('--valid_file', help='Validation file relative name', defaul
 parser.add_argument('--test_file', help='Test file relative name', default='eng.testb.bio')
 parser.add_argument('--suffix', help='Suffix to append', default='.iobes')
 parser.add_argument("--fields", help="The fields to convert", default=[-1], type=int, nargs="+")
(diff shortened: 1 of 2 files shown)


dpressel:
agree with the objective. LGTM

PR #406, opened 2019-12-06
chore stop illegal outputs in the last position of the tf tagger

blester125:
In IOBES tagging there are only a subset of tags that can be the
last tag in a sequence, for example you can't end on a B or I tag
because it will result in a span that doesn't end with an E. In
our constrained decoding we handle this by having a set of illegal
transition from tags to the EOS symbol. In our pytorch CRF the
transition to EOS is calculated inside. In tf we use the built in
CRF which doesn't include transitions to EOS inside the CRF. We
fix that here by inserting EOS tokens explicitly into the emission
scores.

When I implemented warnigs from the iobes chucking code for illegal tags at the end of a sequence I noticed TF had a lot of these warnings while pytorch had none. I dug in and found that this is because there is no transition to EOS considered for the tf tagger.

This change adds the EOS tags into the tensorflow input so that we don't have these illegal moves anymore. I run a lot of taggers for 1 epoch (for speed reasons) comparing the feature/v2 version and this one and after one epoch this version had an average F1 of 88.1 while the old version had 87.9. I haven't run extensive tests on the differences of models trained to convergence.

I also plan to test how this effects speed before we merge but this is only happens at test time so it might not be a big deal if it slows things down.

### DIFF
--- a/python/eight_mile/tf/layers.py
+++ b/python/eight_mile/tf/layers.py
@@ -1493,7 +1493,7 @@ class TaggerGreedyDecoder(tf.keras.layers.Layer):
         if self.inv_mask is not None:
             probv, lengths = TaggerGreedyDecoder.add_go_eos(unary, lengths)
             viterbi, path_scores = crf_decode(probv, self.transitions, lengths)
-            return tf.identity(viterbi[:, 1:], name="best"), path_scores
+            return tf.identity(viterbi[:, 1:-1], name="best"), path_scores
         else:
             return tf.argmax(unary, 2, name="best"), None
 
--- a/python/eight_mile/tf/layers.py
+++ b/python/eight_mile/tf/layers.py
@@ -1414,7 +1414,7 @@ class TaggerGreedyDecoder(tf.keras.layers.Layer):
             _, inv_mask = constraint_mask
             self.inv_mask = inv_mask * tf.constant(-1e4)
 
-            self.A = self.add_weight("transitions_raw", shape=(num_tags, num_tags), dtype=tf.float32, init='zeros', trainable=False)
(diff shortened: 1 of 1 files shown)


blester125:
Based on the results of 10 runs it seems that doing things correctly like this (while eliminating the illegal final token) actually hurts performance right now. Probably something like be eliminating `B-Per`  as the last token something like `S-Loc` is selected instead of `S-Per` like we would want. 

```
$ python analyze.py og fixed
og timing
             time                                   
             mean        std         min         max
phase                                               
Test   145.451812  23.297868  116.299467  181.090452
Train  214.619964  26.548261  166.325820  246.378907
Valid   27.148947   3.457898   20.712076   32.422388
fixed timing
             time                                   
             mean        std         min         max
phase                                               
Test   173.844866   7.367079  162.049975  189.053331
Train  221.030475  12.995983  189.017483  239.200374
Valid   29.365987   1.039741   26.365450   32.718898

og performance
                 value                              
                  mean       std       min       max
phase metric                                        
Test  acc     0.980468  0.000567  0.979385  0.981207
      f1      0.915647  0.002710  0.910853  0.920537
Valid acc     0.987998  0.001506  0.974388  0.989763
      f1      0.945982  0.006359  0.890646  0.953317
fixed performance
                 value                              
                  mean       std       min       max
phase metric                                        
Test  acc     0.980028  0.000678  0.979107  0.980714
      f1      0.914975  0.002431  0.909972  0.917851
Valid acc     0.987927  0.001403  0.976075  0.989433
      f1      0.945783  0.005853  0.897812  0.952317
```

My guess is this has to do with the fact that the GO and EOS transitions are not considered during training so it doesn't learn not to assign high unary probs to illegal tokens. I plan to investigate doing this during training too to see what happens but I am closing this for now.

PR #407, opened 2019-12-06
Feature: add transformer masked LM

wenshuoliu:
The PR adds transformer masked LM by just replacing the `subsequent_mask` by a matrix with all 1s. Also some small changes were made to `pretrain-transformer-lm.py` to ensure `mlm` option works properly.


wenshuoliu:
Just realize that this won't work for finetuning when input is padded. Fixing it.

wenshuoliu:
With another look, the mask during finetuning is actually created in `addons/embed_tlm_pytorch.py`. So as long as we don't pad during pretraining, the current version should be fine.

PR #409, opened 2019-12-12
fix: a couple of bugs in mlm

wenshuoliu:
When running `pretrain-transformer-lm.py` with `torch==1.2.0`, one gets hundreds of warnings like 
```UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.```
which overwhelm the job log. This PR fixes it.


wenshuoliu:
there's (at least) another bug, fixing it

wenshuoliu:
All bugs are fixed (to me). To summarize, this PR:
- removes the warnings `indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead`. 
- fixed the bug in masking indices when training mlm. 
- fixes the bug of different numbers of parameters for lm and mlm (due to a typo). 
- improves the implementation of mlm.

PR #414, opened 2019-12-18
Fix/pickle file imported file

blester125:
Before when we import directly from file we couldn't import it by name later, this cased errors when we tried to pickle addon vectorizers because the module name (which is part of the vectorizer class) wasn't in `sys.modules` so it tried to find the module in sys.path but couldn't because we directly imported from the file.

This change adds the module to `sys.modules` so that later imports of it will work and we can pickle vectorizers in files that are loaded directly.

One question is about how to handle models created with these kind of addons. Normally when we have an add on that includes a vectorizer the module name is saved into `vectorizers-{pid}.json` and when we reload vectorizers we first import each module in the json file so that we can reload the pickle file. In this case the module name that is saved is just the name, not the full file path so when we try to import that it fails because it isn't in the python path. Saving the full file path seems bad because it would tie a model directly to one users setup. It seems like if you have an addon that is used via path over name you need to have a `--modules` argument for things like `api-examples/classify-text.py` or else add the addon to your python path when running it.


sagnik:
I had some discussions with brian and I think I understand the change and I am fine with it.

PR #425, opened 2020-01-15
Feature/relative attention

dpressel:
Implements: https://www.aclweb.org/anthology/N18-2074.pdf

### DIFF
--- a/python/eight_mile/pytorch/embeddings.py
+++ b/python/eight_mile/pytorch/embeddings.py
@@ -14,6 +14,7 @@ from eight_mile.pytorch.layers import (
     WithDropout,
     unsort_batch,
     pytorch_linear,
+    TransformerEncoderStackWithLengths
 )
 
 
@@ -162,6 +163,41 @@ class CharLSTMEmbeddings(PyTorchEmbeddings):
 
     def get_vsz(self):
         return self.embed.get_vsz()
+
+
+class CharTransformerEmbeddings(PyTorchEmbeddings):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.embed = LookupTableEmbeddings(**kwargs)
+        self.d_model = kwargs.get('wsz', 30)
+        self.num_heads = kwargs.get('num_heads', 3)
+        self.rpr_k = kwargs.get('rpr_k', 10)
+        layers = kwargs.get('layers', 1)
+        pdrop = kwargs.get('pdrop', 0.5)
+        self.char_comp = TransformerEncoderStackWithLengths(self.num_heads, self.d_model, pdrop, False, layers,
+                                                            rpr_k=self.rpr_k, input_sz=self.embed.output_dim)
+
+    def forward(self, xch):
+        B, T, W = xch.shape
+        flat_chars = xch.view(-1, W)
+        char_embeds = self.embed(flat_chars)
+
+        # Calculate the lengths of each word
+        lengths = torch.sum(flat_chars != Offsets.PAD, dim=1)
+        patched_lengths = lengths.masked_fill(lengths == 0, 1)
+
+        results = self.char_comp((char_embeds, patched_lengths))
+        # B,T,H output, how to pool this
+        pooled = torch.max(results, -2, keepdims=False)[0]
+        return pooled.reshape((B, T, -1))
+
+
+    def get_dsz(self):
+        return self.d_model
+
+    def get_vsz(self):
+        return self.embed.get_vsz()
+
 
 
 class PositionalMixin(nn.Module):
--- a/python/eight_mile/pytorch/embeddings.py
+++ b/python/eight_mile/pytorch/embeddings.py
@@ -184,9 +184,7 @@ class CharTransformerEmbeddings(PyTorchEmbeddings):
 
         # Calculate the lengths of each word
         lengths = torch.sum(flat_chars != Offsets.PAD, dim=1)
-        patched_lengths = lengths.masked_fill(lengths == 0, 1)
-
-        results = self.char_comp((char_embeds, patched_lengths))
+        results = self.char_comp((char_embeds, lengths))
         # B,T,H output, how to pool this
         pooled = torch.max(results, -2, keepdims=False)[0]
         return pooled.reshape((B, T, -1))
--- a/python/eight_mile/pytorch/layers.py
+++ b/python/eight_mile/pytorch/layers.py
@@ -1942,6 +1942,8 @@ class MultiHeadedRelativeAttention(nn.Module):
         self.attn = None
 
     def make_rpr(self, seq_len, device) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Create a matrix shifted by self.rpr_k and bounded between 0 and 2*self.rpr_k to provide 0-based indexing for embedding
+        """
         seq = torch.arange(seq_len).to(device)
         window_len = 2*self.rpr_k
         edges = seq.view(1, -1) - seq.view(-1, 1) + self.rpr_k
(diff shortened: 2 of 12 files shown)


python/eight_mile/pytorch/embeddings.py
```diff
@@ -164,6 +165,41 @@ def get_vsz(self):
...
+                                                            rpr_k=self.rpr_k, input_sz=self.embed.output_dim)
+
+    def forward(self, xch):
+        B, T, W = xch.shape
+        flat_chars = xch.view(-1, W)
+        char_embeds = self.embed(flat_chars)
+
+        # Calculate the lengths of each word
+        lengths = torch.sum(flat_chars != Offsets.PAD, dim=1)
+        patched_lengths = lengths.masked_fill(lengths == 0, 1)
```
blester125: Do we need to patch the lengths? This is done in the LSTM to avoid cuda errors, Does the transformer error with a length of `0`? It seems like it should be able to handle it, the mask will just have a row of all zeros which seems fine

python/eight_mile/pytorch/layers.py
```diff
@@ -1718,6 +1720,91 @@ def _attention(self, query: torch.Tensor, key: torch.Tensor, mask: Optional[torc
...
+    TODO: where to form RPR?
+    """
+    def __init__(self, hsz: int = None, pdrop: float = 0.1, **kwargs):
+        super().__init__()
+        self.hsz = hsz
+        self.dropout = nn.Dropout(pdrop)
+        self.attn = None
+
+    def forward(self, qkvrm: Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]) -> torch.Tensor:
+        query, key, value, edges_key, edges_value, mask = qkvrm
```
blester125: The name `qkvrm` and the type hint suggests there are 5 things in the tuple but this unpacks into 6 elements

python/eight_mile/pytorch/layers.py
```diff
@@ -1807,15 +1894,98 @@ def forward(self, qkvm: Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Te
...
+            self.attn_fn = SeqScaledDotProductRelativeAttention(dropout)
+        else:
+            self.attn_fn = SeqDotProductRelativeAttention(dropout)
+        self.attn = None
+
+    def make_rpr(self, seq_len, device) -> Tuple[torch.Tensor, torch.Tensor]:
+        seq = torch.arange(seq_len).to(device)
+        window_len = 2*self.rpr_k
+        edges = seq.view(1, -1) - seq.view(-1, 1) + self.rpr_k
+        edges[edges < 0] = 0
```
blester125:
```python
edges[edges < 0] = 0
edges[edges > window_len] = window_len
```

Can be replaced with

```python
edges = torch.clamp(edges, 0, window_len)
```

python/eight_mile/pytorch/layers.py
```diff
@@ -1807,15 +1894,98 @@ def forward(self, qkvm: Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Te
...
+        self.w_K = Dense(d_model, self.d_k * self.h)
+        self.w_V = Dense(d_model, self.d_k * self.h)
+        self.w_O = Dense(self.d_k * self.h, d_model)
+        if scale:
+            self.attn_fn = SeqScaledDotProductRelativeAttention(dropout)
+        else:
+            self.attn_fn = SeqDotProductRelativeAttention(dropout)
+        self.attn = None
+
+    def make_rpr(self, seq_len, device) -> Tuple[torch.Tensor, torch.Tensor]:
```
blester125: Because the pytorch embeddings don't allow for negative numbers I see that you are shifting all the distances by `self.rpr_k` so that a distance of `-self.rpr_k` becomes `0` so the emebddings can handle it. I did the same thing in my SNLI implementation [here](https://github.com/blester125/baseline/blob/dbfdac7515dd67d37651701a4538a9679048686f/python/addons/NLI.py#L293) so that makes sense but there should be some comments to explain what you are doing. At first glance it looks like we are ignoring the negative positions

python/eight_mile/pytorch/layers.py
```diff
@@ -1718,6 +1720,91 @@ def _attention(self, query: torch.Tensor, key: torch.Tensor, mask: Optional[torc
...
+        super().__init__(pdrop=pdrop, **kwargs)
+
+    def _attention(self, query: torch.Tensor, key: torch.Tensor, edges_key: torch.Tensor, mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+        """Scaled dot product attention, as defined in https://arxiv.org/abs/1706.03762
+
+        We apply the query to the keys to receive our weights via softmax in a series of efficient
+        matrix operations. In the case of self-attntion the key and query are all low order
+        projections of the same input.
+
+        :param query: a query for alignment. Can come from self in case of self-attn or decoder in case of E/D
```
blester125: These doc strings need to be updated with the new params and shapes

python/eight_mile/pytorch/layers.py
```diff
@@ -1718,6 +1720,91 @@ def _attention(self, query: torch.Tensor, key: torch.Tensor, mask: Optional[torc
...
+        :param key: a set of keys from encoder or self
+        :param mask: masking (for destination) to prevent seeing what we shouldnt
+        :param edges_key: a matrix of relative embeddings between each word in a sequence [TxTxD]
+        :return: A tensor that is (BxHxTxT)
+        """
+        # (., H, T, T) = (., H, T, D) x (., H, D, T)
+        B, H, T, d_k = query.shape
+        scores_qk = torch.matmul(query, key.transpose(-2, -1))
```
blester125: Shape comments would be nice here

python/eight_mile/pytorch/layers.py
```diff
@@ -1718,6 +1720,91 @@ def _attention(self, query: torch.Tensor, key: torch.Tensor, mask: Optional[torc
...
+        tbhd = query.reshape(B*H, T, d_k).transpose(0, 1)
+        scores_qek = torch.matmul(tbhd, edges_key.transpose(-2, -1))
```
blester125: edges key is `[T, T, d_k]` transposed to `[T, d_k, T]`
tbhd is `[T, B*H, d_k]`
the mult returns `[T, B*H, T]`
The transpose creates `[B*H, T, T]`
The reshapes gives us `[B, H, T, T]`

This seems to work and match the paper :+1:

python/eight_mile/pytorch/layers.py
```diff
@@ -1718,6 +1720,91 @@ def _attention(self, query: torch.Tensor, key: torch.Tensor, mask: Optional[torc
...
+        and values are all low order projections of the same input.
+
+        :param a: The attention weights [B, H, T, T]
+        :param value: The values [B, H, T, D]
+        :param edge_value: The edge values [T, T, D]
+        :returns: A tensor of shape [B, H, T, D]
+        """
+        B, H, T, D = value.shape
+        updated_values = torch.matmul(a, value)
+        a = a.view(B*H, T, T).transpose(0, 1)  # (T, BxH, T)
```
blester125: This reshape seems much cleaner than the ones they do in t2t

```
x_t = tf.transpose(x, [2, 0, 1, 3])
# x_t_r is [length or 1, batch_size * heads, length or depth]
x_t_r = tf.reshape(x_t, [length, heads * batch_size, -1])
```

python/eight_mile/pytorch/layers.py
dpressel: The type hints will show this but I think probably commenting is the best solution.  We could come up with some delimiters in the name, e.g. `q_k_v_rr_rv_m`.

blester125: There are only 5 `torch.Tensor`s in the `Tuple` in the type hint.

PR #426, opened 2020-01-22
disk backed dataset, k8s running, and splitting data

blester125:
This PR does the following:

 * Adds some logging to the data reading (vocab generation and featurization) so we know where things go wrong if there is a crash
 * Fixes a bug where utterances weren't being `.split()`ed when loading the features
 * Fixes a bug where the rank would cause trying to use a gpu you don't have in an environment like k8s where you only ever get gpu `0`
 * Adds a disk backed reader and pytorch dataset object. This lets us avoid crashes I was seeing then trying to read the whole dataset into memory in the `load_features` method
 * Adds a jagged tensor reader. Tensors are stored in a list so they don't have to be padded. there is then a `pad_batch` function used as the `collate_fn` for the `DataLoader` this lets us keep the whole dataset in memory, padded tensors take up too much space because the average utt only takes about 1/4 of the mxlen

**Currently I am not able to get it to train with the all loss, I am looking into this**

### DIFF
--- a/api-examples/pretrain-paired.py
+++ b/api-examples/pretrain-paired.py
@@ -53,6 +53,30 @@ def files_for_dir(directory):
     for file in glob.glob(f'{directory}/*.csv'):
         yield file
 
+
+class SavableFastBPE(object):
+    """Wrapper to define how to pickle fastBPE tokenizer"""
+    def __init__(self, codes_path, vocab_path):
+        from fastBPE import fastBPE
+        self.codes = open(codes_path, 'rb').read()
+        self.vocab = open(vocab_path, 'rb').read()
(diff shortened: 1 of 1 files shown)


api-examples/pretrain-paired.py
```diff
@@ -290,6 +296,145 @@ def load(self, filename, vocabs):
         y_tensor = torch.tensor(features['y'], dtype=torch.long)
         return TensorDataset(x_tensor, y_tensor)
 
+class TensorWordJaggedDatasetReader(TensorWordDatasetReader):
```
dpressel: lets not do this unless there is a really good reason

PR #439, opened 2020-01-24
Fix pretrain transformer lm

wenshuoliu:
This PR made the following changes:
- fix the `proj_to_dsz` bug: in `TransformerLanguageModel`, when the embedding dim is different from the self-attention dim, the model does a projection between them. I put this projection in `create_mask` by mistake in a previous PR. This PR fixes it, and adds this projection in fine-tuning code as well. 
- add `[CLS]` to the vocab when pretraining using `word` or `chars`, and override corresponding vectorizers in the addon to generate `[CLS]` token. 
- add new subclass so that we can fine-tune model pretrained with `chars`.

### DIFF
--- a/python/addons/embed_tlm_pytorch.py
+++ b/python/addons/embed_tlm_pytorch.py
@@ -9,11 +9,85 @@ from baseline.pytorch.transformer import TransformerEncoderStack, subsequent_mas
 from baseline.pytorch.embeddings import PositionalLookupTableEmbeddings
 from baseline.embeddings import register_embeddings
 from baseline.pytorch.embeddings import PyTorchEmbeddings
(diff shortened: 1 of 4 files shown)


python/addons/embed_tlm_pytorch.py
```diff
@@ -231,11 +308,13 @@ def _model_mask(self, nctx):
             return subsequent_mask(nctx)
 
     def forward(self, x):
-        # the following line masks out the attention to padding tokens
+        # the following line masks out the attention to padding tokens: Bx1x1xT
```
blester125: Normally we annotate shapes as `[B, 1, 1, T]` a consistent shape just makes things easier to read

PR #441, opened 2020-01-25
chore: add hook for the output layer of EmbedPoolStackModel

blester125:
This updates the EmbedPoolStackModel and the baseline classify model to allow for overriding the output layer.


dpressel:
Addressed in #442

PR #442, opened 2020-01-26
Feature/layers submodule

dpressel:
This change moves the layers API (`eight_mile`) out of `mead-baseline` and updates to pulling a submodule from `mead-layers`.

To run as before, either `pip install -e .` on the `baseline/layers` sub-module prior to installing baseline, or create a softlink from `layers/eight_mile` into `baseline/python`.


dpressel:
Merge #446 instead

PR #444, opened 2020-02-03
fix: make the EmbeddingsStack have a .keys() like a dict

blester125:
Right now the seq2seq model calls `.keys()` on the embeddings stack object which fails. This adds the `.keys()` call to that object so it doesn't fail.

PR coming to mead-layers soon


dpressel:
This will be handled by merging sub-modules.  Closing for now

PR #453, opened 2020-02-10
Feature/bpe tagger support

dpressel:
- This adds support for BPE-style tagging


blester125:
There needs to be a custom service for these bpe taggers right? You don't have the gold when you are running the tagger at inference time so you need to look at the the actual tokens to undo the bpe right?

dpressel:
Correct, will add this separately

PR #455, opened 2020-02-12
Feat/lm service

blester125:
This PR is tracking the addition of a `score` function to the language models, it can give you the conditional probability of a token given a context or the probability of a sequence.

Tested with the tf word level language model, the pytorch kim and word model. The tf kim model is currently broken but I think this can be merged before that is fixed.

### DIFF
--- a/baseline/services.py
+++ b/baseline/services.py
@@ -400,31 +400,52 @@ class LanguageModelService(Service):
         return super().load(bundle, **kwargs)
 
     def score(self, tokens, **kwargs):
+        """Score tokens with a language model.
+
+        Note:
+            Right now the join probability isn't quite right, it is seeding the model
+            with the first token and the values predicted are the next tokens so it
+            includes a score of only the later tokens as well as the prediction of the
+            next token
+
+            The conditional probability is also a bit wrong because it is being seeded
+            with the value of the first token instead of a <GO> token
+
+        :tokens: A sequence of tokens
+        :prob: Should you get the `joint` score or `conditional` probs
+
+        returns: The score based on the probability type
+        """
         if kwargs.get('preproc', None) is not None:
             logger.warning("Warning: Passing `preproc` to `LanguageModelService.predict` is deprecated.")
         tokens_batch = self.batch_input(tokens)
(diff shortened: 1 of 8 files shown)


baseline/services.py
```diff
@@ -399,6 +403,83 @@ def load(cls, bundle, **kwargs):
...
+        # scores = torch.gather(softmax_tokens, -1, values.unsqueeze(-1))
+        scores = []
+        # The last softmax is what comes after the <EOS> so we don't grab from there
+        for soft, value in zip(softmax_tokens[:, :-1], values):
+            tokens = []
+            for tok, val in zip(soft, value):
+                tokens.append(tok[val].item())
+            scores.append(tokens)
+        scores = np.array(scores)
+        logger.warning("Warning: The `joint` probability isn't well formed right now and should be used with a grain of salt")
```
dpressel: I dont think we need this line

layers/eight_mile/utils.py
```diff
@@ -1548,3 +1548,16 @@ def mime_type(file_name: str) -> str:
...
         return "text/plain"
     return "application/w2v"
+
+
+@export
+def to_numpy(x):
+    if isinstance(x, np.ndarray):
+        return x
+    try:
+        import torch
```
dpressel: can actually test if its an instance of a torch.Tensor right?

blester125: Why do we need to test that? If you are using tf1 then your predict result is already a numpy array and is caught by the first check, if you are using tf2 and you don't have pytorch installed the import causes an error which we catch. We then call `.numpy()` to turn the `tf.Tesor` into a numpy array. If you have pytorch installed but it was a tf2 tensor the `.detach()` throws which we again catch and just us `.numpy()`. If we are using pytorch we use `.detach()` which is defined on Tensors with and without `requires_grad=True` and then we move it to `cpu` before calling `.numpy()`

I guess an explicit check (while keeping the try in case they don't have pytorch installed) is probably simpler to reason when we need to look at this function again

dpressel: it could be a little more generalizable in the future, and more precise about what its trying to accomplish.

PR #457, opened 2020-02-13
Update layers to be more consistent with shapes, add shape tests

dpressel:
First, I added GRUs but only using Keras layers on the TF side. 
These arent that commonly used and just implementing it one way shouldnt really be a compatibility concern.

The PyTorch and TF were different in their output shapes.  PyTorch's RNN output is limited to the max length found in the batch, which may not always be the actual tensor width.

Using Keras LSTM/GRU this was failing because we could generate a sequence mask that was shorter than the tensor width.  In TF 1.x the output would always be the size of the input.

I added some shape/layer tests for common encoders.  This helped find inconsistencies in API.

Also, I found a mismatch in the TF All in 1.x vs Keras.
The legacy was not packaging the hidden layers consistently with the other versions.
This is all right in theory, as we can just use whatever and pass it downstream, but in practice, we were looking for drop in replacement and so this caused things like shape tests to fail.

During discussions on proper behavior for RNNs, we decided to make them consistent with PyTorch,
which means that we now have to trim the inputs according to the max sequence length for all
TF versions.

We also decided that it would be nice to make the conv layers guaranteed to not exceed the temporal length of the input tensor.  This is accounted for by replacing the `padding=kernel_size//2` with a `Conv1DSame` which expands the tensor to the normal pad on the front and  1 px less on the backend of the tensor when its an even kernel.  This makes the output shape guaranteed to match the input size.  Constant zero padding is applied.

The conv encoder stack number of layers param was different from the RNNs and was ordered on the other side of `pdrop` so I changed those to be consistent with the ordering of the RNN stacks.

I also made the default conv encoders `[B, T, H]` with an option `hidden_last=True`.  The encoder stacks only do this on input and output to keep a stack of layers internally consistent as `[B, H, T]` for convolution efficiency.

### DIFF
--- a/layers/eight_mile/tf/layers.py
+++ b/layers/eight_mile/tf/layers.py
@@ -815,16 +815,11 @@ class LSTMEncoderSequence1(LSTMEncoder1):
         return output
 
 
-class LSTMEncoderAll1(LSTMEncoder1):
+class LSTMEncoderAllLegacy(LSTMEncoder1):
     """LSTM encoder that passes along the full output and hidden states for each layer
 
-    Takes a tuple containing a tensor input of shape `[B, T, C]` and lengths of shape `[B]`
-
-    This returns a 2-tuple of outputs `[B, S, H]` where `S = max(lengths)`, for the output vector sequence,
-    and a tuple of hidden vector `[L, B, H]` and context vector `[L, B, H]`, respectively
-
-    *TF Note*: This module reorganizes the underlying TF output, which means you must be careful if using
-    this with another tf 1 `RNNCell` implementation
+    *TF Note*: This module does not change the underlying TF output, we only support this to make it easier to use
+    with existing TF primitives, especially in seq2seq
     """
     def call(self, inputs):
         """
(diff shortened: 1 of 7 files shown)


layers/eight_mile/tf/layers.py
```diff
@@ -848,6 +1094,8 @@ def output_fn(self, rnnout, state):
     def call(self, inputs):
         inputs, lengths = tensor_and_lengths(inputs)
         mask = tf.sequence_mask(lengths)
+        max_length = tf.reduce_max(lengths)
```
blester125: I know this is a pretty easy thing to do (only two lines) but it seems like it will be sprinkled throughout a lot of the code, might make sense as a function?

layers/eight_mile/tf/layers.py
```diff
@@ -848,6 +1094,8 @@ def output_fn(self, rnnout, state):
...
+        inputs = inputs[:, :max_length, :]
```
blester125: This seems overly prescriptive to the `[B, T, H]` shape. Does it make sense for a function that takes a dim and only culls on the dim specified? Right now this would fail at cutting a tensor that was like `[B, T, H, H2]` or something

dpressel: would that make sense for an RNN input though?

layers/eight_mile/tf/layers.py
dpressel: maybe a private one, op is different if (Bi)LSTM*1 or 2

layers/eight_mile/tf/layers.py
blester125: Not right now but one of the pain points in dynet was that was hard to handle things that had extra dimension like the transformer attention so figured we would want to move away from things being so locked into the `[B, T, H]` mindset

dpressel: we can engineer it more when there are less redundant things

layers/eight_mile/tf/layers.py
dpressel: this is a good point, but I think it can be cleaned up later

PR #461, opened 2020-02-13
feat: a safe way to infer tensor lengths

blester125:
Based on the comment here https://github.com/dpressel/mead-baseline/blob/509453992838a524f6442d4e0f07a034390ae1f7/layers/eight_mile/pytorch/layers.py#L70

This adds functions that infer the length of an example but they are some what safer than just summing the number of non `<PAD>` elements because they find the last non padded element and us that +1 as the length

I didn't actually make the change to always return the lengths tho because I didn't know what repercussions that would have


dpressel:
looks good

PR #471, opened 2020-02-27
Feature/load learned positional

wenshuoliu:
This PR updates saver, loader and addons for tlm using learned-positional.

### DIFF
--- a/layers/eight_mile/pytorch/serialize.py
+++ b/layers/eight_mile/pytorch/serialize.py
@@ -117,25 +117,35 @@ def from_encoder_array(pytorch_encoder: TransformerEncoder, d: Dict, name: str):
 
 
 def to_embed_array(pytorch_embed: nn.Module, name: str) -> Dict:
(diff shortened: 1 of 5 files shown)


layers/eight_mile/pytorch/serialize.py
```diff
@@ -212,7 +220,7 @@ def from_tlm_array(pytorch_tlm: nn.Module, d: Dict, embeddings_key: str = 'x', n
     :return:
     """
     from_encoder_stack_array(pytorch_tlm.transformer, d, name=f"{name}/TransformerEncoderStack")
-    from_embed_array(pytorch_tlm.embeddings[embeddings_key].embeddings, d, f"{name}/SinusoidalPositionalEmbeddings")
+    from_embed_array(pytorch_tlm.embeddings[embeddings_key], d, f"{name}/PositionalEmbeddings")
```
dpressel: this change will make all previous checkpoints incompatible

PR #474, opened 2020-03-03
Transformer paired pre-training and out-of-core datasets

dpressel:
- Creates a `transformer_utils.py` that contains the loss functions and model for ConveRT as well as useful pytorch `Datasets`s, including a streaming, multiworker dataset that can be used for reddit
- Support for dual encoders and seq2seq transformers with `pretrain-paired.py` script quite a bit under the constraints of processing a massive dataset
- Adds an LM for out-of-core processing on single line data
- Reader for reddit supports Next Turn Prediction, Next Sequence Prediction and LM Prediction
- Fixes a bug in the scaling for the dual encoder where it shouldve been using `sqrt(length)`


api-examples/transformer_utils.py
```diff
@@ -0,0 +1,422 @@
...
+        P(y|x) = e^S(x, y) / \sum_i^k e^S(x, y_k)
+        log P(y|x) = log( e^S(x, y) / \sum_i^k e^S(x, y_k))
+        log P(y|x) = S(x, y) - log \sum_i^k e^S(x, y_k)
+        -log P(y|x) = -(S(x, y) - log \sum_i^k e^S(x, y_k))
+        """
+        super().__init__()
+        self.c = 1.0
+        self.score = nn.CosineSimilarity(dim=-1)
+        self.model = model
+        self.max_scale = math.sqrt(self.model.embedding_layers.get_dsz())
```
blester125: From the paper it says `computed as cosine similarity with annealing between the encodings hx and hy. It starts at 1 and ends at √d` so it seems like this should actually be the sqrt of the size of the inputs rather than the size of the embeddings. I know that our transformers don't really change sizes as data flows trough but I think this seems unsafe

dpressel: the input size in this code is guaranteed to be the same between embeddings and hidden units.

api-examples/transformer_utils.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def __len__(self):
+        return self.samples
+
+    def __iter__(self):
+        # Each node has the same worker_info, so the unique offsets for each is
+        # rank * num_workers + worker_id
+        # and the total available workers is world_size * num_workers
+        worker_info = torch.utils.data.get_worker_info()
+        files = list(glob.glob(f"{self.directory}/{self.pattern}"))
```
blester125: Does `glob.glob` return a deterministic ordering? If it doesn't (for example last mod time could change where a file turns up in the list) we could have two workers who read the files in different orders which could result in reading the same files with different workers and probably missing some other files.

It seems like it returns the same order of `ls -U` but I don't know if that changes. It might be safer (but a little slow)  to sort this list

api-examples/transformer_utils.py
```diff
@@ -0,0 +1,422 @@
...
+        if reader_type == "ntp":
+            return NextTurnPredictionFileLoader(directory, self.pattern, vocabs, self.vectorizer, self.nctx)
+        elif reader_type == "nsp":
+            return NextSequencePredictionFileLoader(directory, self.pattern, vocabs, self.vectorizer, 2*self.nctx)
+        else:
+            print("Using files as an LM")
+            return SequencePredictionFileLoader(directory, self.pattern, vocabs, self.vectorizer, self.nctx)
+
+
+class TiedSeq2SeqModel(Seq2SeqModel):
```
blester125: What is tied here? just the embeddings? The name makes me think it is doing weight sharing between the Encoder and Decoder but I don't see it forcing that unless I am missing something

dpressel: yes, this could have been clearer, it differs from the base class in that it ties the embeddings

PR #484, opened 2020-03-13
feat: Added tools and models for calibration

blester125:
Calibration is an important tool for trustworthy models. When you have a
classifier it often outputs scores for each class that people
interpret as probabilities for each class. The problem is these are
often not well calibrated. This means that when you have 100 examples
that all have a score of 0.8 we would expected that about 80 of those
100 examples would be correctly labeled by the model, however; with
neural networks this is often not the case. When a model says that
examples have a probability of 0.8 the proportion of examples that are
correct will be much less than than the expected 80/100. These models
are over confident in their answers.

This PR adds metrics, models, and training for calibration as discussed
in [On Calibration of Modern Neural Networks](https://arxiv.org/abs/1706.04599).
It adds metrics and visualization as an `eight_mile.calibration`
package, adds various post-hoc methods for calibration as addons, adds
and api-example to collect metrics on calibrations, and adds a script
that can combine the various graphs that describe calibration into a
single figure.

The way this models are trained is with a checkpoint. First you train a
model as normal with `mead-train` and then you use a second config that
specifies the same model but set the model type to one of the new
calibrations methods and pass the checkpoint of the old model in. You
need to make sure the model is specified the same as the old model or
else weights won't get read into the checkpoint.

The final thing this PR does it that it updates the classify trainer to
only update the `.trainable_variables` of a model. This means that by
controlling this value in the post-hoc calibration models we can in
effect freeze the rest of the network weights by ignoring them in the
optimization.

### DIFF
--- a/baseline/tf/classify/training/utils.py
+++ b/baseline/tf/classify/training/utils.py
@@ -101,16 +101,14 @@ class ClassifyTrainerTf(EpochReportingTrainer):
 
         """
         super().__init__()
-
         if type(model_params) is dict:
             self.model = create_model_for('classify', **model_params)
         else:
             self.model = model_params
-
         self.sess = self.model.sess
         self.loss = self.model.create_loss()
         self.test_loss = self.model.create_test_loss()
-        self.global_step, train_op = optimizer(self.loss, colocate_gradients_with_ops=True, **kwargs)
+        self.global_step, train_op = optimizer(self.loss, colocate_gradients_with_ops=True, variables=self.model.trainable_variables, **kwargs)
         self.nsteps = kwargs.get('nsteps', six.MAXSIZE)
         decay = kwargs.get('ema_decay', None)
         if decay is not None:
@@ -218,12 +216,13 @@ class ClassifyTrainerTf(EpochReportingTrainer):
             y = batch_dict['y']
             if use_dataset:
                 guess, lossv = self.sess.run([self.model.best, self.test_loss])
+                batchsz = len(guess)
             else:
                 feed_dict = self.model.make_input(batch_dict, False)
                 guess, lossv = self.sess.run([self.model.best, self.test_loss], feed_dict=feed_dict)
 
-            batchsz = self._get_batchsz(batch_dict)
-            assert len(guess) == batchsz
+                batchsz = self._get_batchsz(batch_dict)
+                assert len(guess) == batchsz
             total_loss += lossv * batchsz
             total_norm += batchsz
             cm.add_batch(y, guess)
--- a/baseline/tf/classify/training/utils.py
+++ b/baseline/tf/classify/training/utils.py
@@ -129,10 +129,6 @@ class ClassifyTrainerTf(EpochReportingTrainer):
             skip_blocks = kwargs.get('blocks_to_skip', ['OptimizeLoss'])
             reload_checkpoint(self.model.sess, checkpoint, skip_blocks)
 
-    @staticmethod
-    def _get_batchsz(batch_dict):
-        return len(batch_dict['y'])
-
     def _train(self, loader, dataset=True, **kwargs):
         """Train an epoch of data using either the input loader or using `tf.dataset`
 
@@ -216,13 +212,11 @@ class ClassifyTrainerTf(EpochReportingTrainer):
             y = batch_dict['y']
             if use_dataset:
                 guess, lossv = self.sess.run([self.model.best, self.test_loss])
-                batchsz = len(guess)
             else:
                 feed_dict = self.model.make_input(batch_dict, False)
                 guess, lossv = self.sess.run([self.model.best, self.test_loss], feed_dict=feed_dict)
 
-                batchsz = self._get_batchsz(batch_dict)
-                assert len(guess) == batchsz
+            batchsz = len(guess)
             total_loss += lossv * batchsz
             total_norm += batchsz
             cm.add_batch(y, guess)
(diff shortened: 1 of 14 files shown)


baseline/tf/classify/training/utils.py
```diff
@@ -101,16 +101,14 @@ def __init__(self, model_params, **kwargs):
...
         if type(model_params) is dict:
             self.model = create_model_for('classify', **model_params)
         else:
             self.model = model_params
-
         self.sess = self.model.sess
         self.loss = self.model.create_loss()
         self.test_loss = self.model.create_test_loss()
-        self.global_step, train_op = optimizer(self.loss, colocate_gradients_with_ops=True, **kwargs)
+        self.global_step, train_op = optimizer(self.loss, colocate_gradients_with_ops=True, variables=self.model.trainable_variables, **kwargs)
```
dpressel: should this idiom be applied more broadly?  other tasks?

layers/eight_mile/calibration/calibration.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    preds = np.argmax(probs, axis=1)
+    pred_probs = np.max(probs, axis=1)
+    return binary_calibration_bins(true == preds, pred_probs, bins=bins)
+
+
+calibration_bins = multiclass_calibration_bins
+
+
+def binary_calibration_bins(true: np.ndarray, probs: np.ndarray, bins: int) -> Bins:
```
dpressel: correct?

blester125: That makes more sense. I'll make that change soon

blester125:
From some discussion we think that there would be a weighting parameter to the binning functions. This would allow for users to weight different classes depending on how important it is that class is calibrated.

layers/eight_mile/calibration/calibration.py
blester125: Updated

layers/eight_mile/calibration/calibration.py
```diff
@@ -0,0 +1,73 @@
...
+    :param probs: A 1D array of the probabilities from a model. Each row represents an example in the dataset.
+        and each column represents the probably assigned by the model to each class for that example.
+    :param bins: The number of bins to use when aggregating.
+    :param class_weights: A 1D array of scores that can add extra weight to examples of specific classes.
+
+    :returns: The metrics aggregated by bins
+    """
+    preds = np.argmax(probs, axis=1)
+    pred_probs = np.max(probs, axis=1)
+    correct = true == preds
```
dpressel: I think we should rename this param.

blester125: Any idea to what? I already changed to to correct like you mentioned before, I guess it is a little odd now that it can have a score on it

dpressel: I mean `true`

blester125: `gold`? This parameter is a list of the true labels so I figured it made sense?

blester125: updated

baseline/tf/classify/training/utils.py
blester125: I think it should be done but I would rather get this in and track that in another merge because there is probably a lot of testing that needs to be done for that https://github.com/dpressel/mead-baseline/issues/493

layers/eight_mile/calibration/calibration.py
```diff
@@ -0,0 +1,74 @@
...
+__all__ = ["calibration_bins", "multiclass_calibration_bins", "binary_calibration_bins", "average_confidence", "Bins"]
+
+
+Bins = namedtuple("Bins", "accs confs counts edges")
+
+
+def multiclass_calibration_bins(truth: np.ndarray, probs: np.ndarray, bins: int, class_weights: Optional[np.ndarray] = None) -> Bins:
+    """Calculate the binned confidence and accuracy for a multiclass problem.
+
+    :param true: A 1D array of the true labels for some examples.
```
dpressel: param name changed but pydoc isnt updated

layers/eight_mile/calibration/calibration.py
```diff
@@ -0,0 +1,74 @@
...
+    bin_idx = np.digitize(probs, bins) - 1
+
+    bin_conf_sum = np.bincount(bin_idx, weights=probs, minlength=len(bins))
+    bin_acc_sum = np.bincount(bin_idx, weights=credit, minlength=len(bins))
+    bin_counts = np.bincount(bin_idx, minlength=len(bins))
+
+    mask = bin_counts != 0
+    denom = np.where(mask, bin_counts, 1)
+
+    bin_mean_conf = (bin_conf_sum / denom) * mask
```
dpressel: If the bin counts are 0, how is the numerator 0?  Seems like the mask multiply will produce the same result as without it

dpressel: meant to say how is numerator *not* 0

layers/eight_mile/calibration/metrics/calibration_error.py
```diff
@@ -0,0 +1,54 @@
...
+    bin_mean_conf: np.ndarray,
+    bin_counts: np.ndarray,
+) -> float:
+    """The difference in expectation between the confidence and the accuracy.
+
+    This is an approximation of eq (2) as described in eq (5)
+        from https://arxiv.org/abs/1706.04599
+
+    This is the absolute difference between the confidence and accuracy of each
+    bin weighted but the number of samples in each bin.
```
dpressel: by

layers/eight_mile/calibration/calibration.py
blester125: you right, removed

layers/eight_mile/calibration/calibration.py
blester125: fixed

layers/eight_mile/calibration/metrics/calibration_error.py
blester125: fixed

PR #485, opened 2020-03-13
use pytorch data loader to train

dpressel:
Use PyTorch's built-in `DataLoader` to traverse the MEAD-generated map-style dataset.
This is a pre-cursor for doing larger-scale processing in the future with access to multiple workers.

Analysis of the timing logs shows that the impact is negligable, and the model performance is not impacted.

One change that is required is to not assume that `make_input` should be accessing numpy arrays,
I added a `numpy_to_tensor` boolean to the argument list and to the sub-methods it calls for each task, and default this to false, so that for all training we do not have it on.

I also made the predict methods support `kwargs` and default to `numpy_to_tensor` on.

### DIFF
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -118,10 +118,10 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
 
     def predict_batch(self, batch_dict, **kwargs):
         numpy_to_tensor = bool(kwargs.get('numpy_to_tensor', True))
-        examples, prem_idx = self.make_input(batch_dict, perm=True, numpy_to_tensor=numpy_to_tensor)
+        examples, perm_idx = self.make_input(batch_dict, perm=True, numpy_to_tensor=numpy_to_tensor)
         with torch.no_grad():
             probs = self(examples).exp()
-            probs = unsort_batch(probs, prem_idx)
+            probs = unsort_batch(probs, perm_idx)
         return probs
 
     def predict(self, batch_dict, raw=False, dense=False, **kwargs):
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -54,54 +54,76 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
     def create_loss(self):
         return nn.NLLLoss()
 
-    def make_input(self, batch_dict, numpy_to_tensor=False):
+    def make_input(self, batch_dict, perm=False, numpy_to_tensor=False):
         """Transform a `batch_dict` into something usable in this model
 
         :param batch_dict: (``dict``) A dictionary containing all inputs to the embeddings for this model
         :return:
         """
         example_dict = dict({})
-        for key in self.embeddings.keys():
-            if numpy_to_tensor:
-                example_dict[key] = torch.from_numpy(batch_dict[key])
-            else:
-                example_dict[key] = batch_dict[key]
-            if self.gpu:
-                example_dict[key] = example_dict[key].cuda()
+        perm_idx = None
 
         # Allow us to track a length, which is needed for BLSTMs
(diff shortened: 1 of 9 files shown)


baseline/pytorch/classify/model.py
```diff
@@ -54,46 +54,76 @@ def cuda(self, device=None):
...
+        perm_idx = None
 
         # Allow us to track a length, which is needed for BLSTMs
         if self.lengths_key is not None:
-            example_dict['lengths'] = torch.from_numpy(batch_dict[self.lengths_key])
+            lengths = batch_dict[self.lengths_key]
+
+            if numpy_to_tensor:
+                lengths = torch.from_numpy(lengths)
+                lengths, perm_idx = lengths.sort(0, descending=True)
```
blester125: This part should be outside of this if right? We can still have the situation where we want to sort the lengths even if using a dataset right?

baseline/pytorch/lm/train.py
```diff
@@ -46,15 +47,16 @@ def _get_pytorch_model(self):
         return self.model.module if self.gpus > 1 else self.model
 
     @staticmethod
-    def _get_dims(batch_dict):
+    def _get_dims(loader):
```
blester125: Not related to this PR because the problem is already there but some of the tf trainers call this `_get_batchsz` we should be consistent

blester125: nvm I see now that that it can be used for dims other than batch size like when getting the number of tokens

baseline/pytorch/lm/train.py
```diff
@@ -46,15 +47,16 @@ def _get_pytorch_model(self):
...
+        batch_dict = loader.dataset[0]
         return batch_dict['y'].shape
 
     @staticmethod
     def _num_toks(batch_dict):
-        return np.prod(LanguageModelTrainerPyTorch._get_dims(batch_dict))
+        return np.prod(batch_dict['y'].shape)
```
blester125: I see that we don't use the `_get_dims` anymore because this `_num_toks` need to be dynamic (needs to report for each batch because it can be different) but that raises the question of if we can remove the `_get_dims` function.

Also although we always uses dense LM batches would it make sense to calculate the number of tokens based on the lengths in case we ever need a LM that has padding (like each example if a single sentence)

baseline/pytorch/classify/model.py
dpressel: yep

baseline/pytorch/seq2seq/train.py
```diff
@@ -43,7 +44,7 @@ def __init__(self, model, **kwargs):
 
     @staticmethod
     def _num_toks(tgt_lens):
-        return np.sum(tgt_lens)
+        return float(torch.sum(tgt_lens))
```
blester125: I thought the suggested way to get the value from a pytorch tensor into a python number was to use `.item`?

Also the number of tokens seems like it should be an int not a float?

baseline/pytorch/classify/model.py
dpressel: fixed

baseline/pytorch/seq2seq/train.py
dpressel: We could.  AFAICT its only ever used to divide out tokens or multiply by another float

baseline/pytorch/lm/train.py
dpressel: We currently use `_get_dims` in cases where we are doing truncated backprop, and its a bit unfortunate that we get the batch size this way as it makes the code clunky and not very robust -- it ends up assuming that we have a map-style dataset which we dont want to assume probably moving forward.

baseline/pytorch/seq2seq/train.py
dpressel: switched to item

PR #487, opened 2020-03-15
chore: add hook for the output layer of EmbedPoolStackModel

blester125:
In the shuffle of deciding if we should move eight_mile into it's own repo the changes that let you override the output hook in the pytorch classifier model got lost. This PR changes them back. It also updates the name if the output from `output_layer` to `output_model` to be consistent with the naming of the other parts of the model (`pool_model`, etc)

### DIFF
--- a/layers/eight_mile/pytorch/layers.py
+++ b/layers/eight_mile/pytorch/layers.py
@@ -1793,10 +1793,9 @@ class EmbedPoolStackModel(nn.Module):
         super().__init__()
         self.embed_model = embeddings
         self.pool_model = pool_model
-        output_dim = self.pool_model.output_dim if stack_model is None else stack_model.output_dim
-        self.output_layer = Dense(output_dim, nc, activation="log_softmax")
         self.stack_model = stack_model if stack_model else nn.Identity()
-        self.output_layer = Dense(output_dim, nc, activation="log_softmax") if output_model is None else output_model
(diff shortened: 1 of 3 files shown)


layers/eight_mile/pytorch/layers.py
```diff
@@ -1793,18 +1793,17 @@ def __init__(
         super().__init__()
         self.embed_model = embeddings
         self.pool_model = pool_model
-        output_dim = self.pool_model.output_dim if stack_model is None else stack_model.output_dim
-        self.output_layer = Dense(output_dim, nc, activation="log_softmax")
         self.stack_model = stack_model if stack_model else nn.Identity()
-        self.output_layer = Dense(output_dim, nc, activation="log_softmax") if output_model is None else output_model
+        output_dim = self.pool_model.output_dim if stack_model is None else stack_model.output_dim
+        self.output_model = Dense(output_dim, nc, activation="log_softmax") if output_model is None else output_model
```
dpressel: I suggest we leave the name as it was before `output_layer`.  The thought is that 1. its less likely to cause broken compat, and that 2. the output is much more likely to be a layer than a model (unlike the others)

blester125: Reverted. It is still a little strange because the argument name from before was `output_model` and it is being saved to an attribute called `output_layer` but whatever

PR #490, opened 2020-03-18
Checkpoint ups for pretraining + example seq2seq generate

dpressel:
- adds a sample program for using pretrained paired seq2seq to do generation
- make the checkpoint type explicit in naming
- allow reloading from a checkpoint of step or epoch type
- suppport legacy checkpoint reloading with `--reload_tt (step|epoch)` option


api-examples/pretrain-transformer-lm.py
```diff
@@ -526,10 +527,25 @@ def train():
     start_epoch = 0
     if args.restart_from:
         model.load_state_dict(torch.load(args.restart_from))
```
wenshuoliu: Speaking of legacy ckpts, when i load the ones created by v1 and converted to v2, this line causes error because those ckpts don't have the overall layernorm. We probably want `strict=False` and a warning if there's mismatch of keys.

api-examples/pretrain-lm-streaming.py
```diff
@@ -209,10 +210,24 @@ def train():
...
+        vec = args.restart_from.split("-")
+
+        if args.restart_tt:
+            tick_type = args.restart_tt
+        else:
+            tick_type = vec[-2]
+        step_num = int(vec[-1].split(".")[0])
+        if tick_type == 'epoch':
+            start_epoch = step_num - 1
+            global_step = (start_epoch + 1) * steps_per_epoch
```
wenshuoliu: In all the cases, `steps_per_epoch` are calculated as `number_of_train_samples // args.batch_size`, but i think it should be `number_of_train_samples // (args.batch_size*num_gpus)` since each worker only needs to run this many steps to finish a epoch. I cut my release and tested that `int(os.environ.get("WORLD_SIZE", 1))` does give the num_gpus in our cluster environment. I still think we should fix this so the scheduler works as expected.

api-examples/pretrain-transformer-lm.py
```diff
@@ -381,6 +381,7 @@ def train():
     parser.add_argument("--weight_decay", type=float, default=0.0, help="Weight decay")
     parser.add_argument("--epochs", type=int, default=20, help="Num training epochs")
     parser.add_argument("--restart_from", type=str, help="Option allows you to restart from a previous checkpoint")
+    parser.add_argument("--restart_tt", type=str, help="Optional param for legacy checkpoints (step|epoch)")
     parser.add_argument("--warmup_steps", type=int, default=1000, help="Num warmup steps")
     parser.add_argument("--mlm", type=str2bool, default=False, help="Use Masked Language Model (MLM) objective")
     parser.add_argument('--rpr_k', help='Relative attention positional sizes pass 0 if you dont want relative attention',
```
wenshuoliu: Off the topic, but we still need lines like `if len(args.rpr_k) == 0 or args.rpr_k[0] < 1: args.rpr_k = None` in this file, or it'll complain.

api-examples/pretrain-transformer-lm.py
dpressel: Lets handle this in a separate PR

api-examples/pretrain-lm-streaming.py
dpressel: Can this be resolved?

api-examples/pretrain-transformer-lm.py
dpressel: Right, but I didnt remove that, missing the issue, can you clarify?

wenshuoliu: It is in `pretrain-paired.py` and `pretrain-lm-streaming.py`, but it's missing in `pretrain-transformer-lm.py`.

dpressel: fixed

PR #498, opened 2020-03-27
Adds class name cascading inside the tagger

dpressel:
Previously, when we called the tagger, we were calling
`.transduce()` followed by `.decode()` so that we could grab
the probabilities from `.transduce().`  This meant that we
werent actually calling the `.call()` method, which caused
us not to scope the name of the models as we do in the
classifier.  This change makes the underlying `.call()` function
take an boolean to support returning the LSTM layer output
optionally, which means that we can now use the proper method
so that Keras/TF can figure out how to properly name our sub-objs

Previous behavior:
```
<tf.Variable 'embeddings_stack/word/word/LUT/Weight:0' shape=(7534, 200) dtype=float32>

```
New behavior:
```
<tf.Variable 'rnn_tagger_model/tag_sequence_model/embeddings_stack/word/word/LUT/Weight:0' shape=(7534, 200) dtype=float32>


```

### DIFF
--- a/baseline/tf/tagger/model.py
+++ b/baseline/tf/tagger/model.py
@@ -18,10 +18,10 @@ class TaggerModelBase(tf.keras.Model, TaggerModel):
     This class provides the implementation of the TaggerModel for TensorFlow
     """
 
-    def __init__(self):
+    def __init__(self, name=None):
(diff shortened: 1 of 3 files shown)


dpressel:
Closing, #542 addresses this better, I think

PR #499, opened 2020-03-27
Fix/tagger name cascading without sub-model

dpressel:
Previously, when we called the tagger, we were calling
.transduce() followed by .decode() so that we could grab
the probabilities from .transduce(). This meant that we
werent actually calling the .call() method, which caused
us not to scope the name of the models as we do in the
classifier. This change makes the TaggerModelBase .call() function
take an boolean to support returning the LSTM layer output
optionally, which means that all sub-classes names will prepend to the graph names:

Previous behavior:
```
<tf.Variable 'embeddings_stack/word/word/LUT/Weight:0' shape=(7534, 200) dtype=float32>

```
Current behavior:
```
<tf.Variable 'rnn_tagger_model/embeddings_stack/word/word/LUT/Weight:0' shape=(7534, 200) dtype=float32>
```

### DIFF
--- a/baseline/tf/tagger/model.py
+++ b/baseline/tf/tagger/model.py
@@ -162,8 +162,10 @@ class TaggerModelBase(tf.keras.Model, TaggerModel):
 
         return batch_for_model
 
-    def call(self, *args, **kwargs):
-        return self.impl(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


dpressel:
Closing, #542 addresses this better, I think

PR #501, opened 2020-03-30
fine-grain control over embedding placement

dpressel:
Previously, we were explicitly forcing all
LUT embeddings onto the CPU in eager mode.
This is extremely bad for performance for optimizers
than can support GPU placed LUT updates and causes
issues with multi-worker distribution

This change adds an explicit variable that can
be passed via config and in cases where default
placement is not possible, will explicitly scope
to CPU device by checking the optimizer and seeing
if we are in eager mode

### DIFF
--- a/baseline/tf/tagger/training/eager.py
+++ b/baseline/tf/tagger/training/eager.py
@@ -5,7 +5,7 @@ import numpy as np
 import tensorflow as tf
 import logging
 from eight_mile.utils import listify, revlut, to_spans, write_sentence_conll, per_entity_f1, span_f1, conlleval_output
-from eight_mile.tf.layers import TRAIN_FLAG, SET_TRAIN_FLAG, reload_checkpoint
+from eight_mile.tf.layers import TRAIN_FLAG, SET_TRAIN_FLAG, reload_checkpoint, get_shape_as_list
 from eight_mile.tf.optz import EagerOptimizer
 from baseline.progress import create_progress_bar
 from baseline.model import create_model_for
@@ -195,7 +195,7 @@ class TaggerTrainerEagerTf(EpochReportingTrainer):
 
             lossv = self.optimizer.update(self.model.impl, features, y).numpy()
             step += 1
-            batchsz = y.shape[0]
+            batchsz = get_shape_as_list(y)[0]
             report_loss = lossv * batchsz
             epoch_loss += report_loss
             epoch_div += batchsz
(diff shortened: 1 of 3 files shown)


dpressel:
*Classification Benchmarking on this branch vs master*

In the case of `adam` optimizer, this branch (`now`) will now use default placement to the GPU in eager mode.
Previously, it would force to the CPU.  For comparisons of `adadelta` the behavior is unchanged from before,
so even though the `branch` is marked as `now`, this would have the same performance stats on the `master` branch.

Interestingly, something changed in declaractive (`decl`) mode between TF 1.x and TF 2.x and it is now MUCH
faster for `adadelta` -- essentially identical speed to `adam`!

tf-version | mode  | optim | task        | dataset | branch | epoch time |
-------------------------------------------------------------------------|
2.1        | eager | adam     | classify |    sst2 | master | 81.8       |
2.1        | eager | adam     | classify |    sst2 | now    | 27.4       |
1.15       | eager | adam     | classify |    sst2 | master | 87.8       |
1.15       | eager | adam     | classify |    sst2 | now    | 28.6       |
1.15       | decl  | adadelta | classify |    sst2 | now    | 59.3       |
1.15       | eager | adadelta | classify |    sst2 | now    | 43.7       |
2.1        | decl  | adadelta | classify |    sst2 | now    | 18.5       |
2.1        | eager | adadelta | classify |    sst2 | now    | 42.8       |

dpressel:
*Tagging Benchmarking on this branch vs master*

In the case of `sgd` optimizer with no momentum, this branch (`now`) will now use default placement to the GPU in eager mode.
Previously, it would force to the CPU.  Since declarative mode doesnt change between versions, I record only the results for the current `branch` as `now`.

Interestingly, right now the SGD behavior on TF 1.x appears particulary
ineffecient in eager mode, and default device placement doesnt help at all.
SGD behavior and is much improved on TF 2.x, and we see a large performance
gain in eager mode on this branch:

--------------------------------------------------------------------------------------------
tf-version | mode  | optim  | task      | dataset        | branch | epoch time |
-------------|----------|----------|-----------|------------------|-----------|----------------|
1.15         | eager | sgd      | tagger  |  conll2003    | master | 544.9         |
1.15         | eager | sgd      | tagger  |  conll2003    | now     | 560            |
2.1           | eager | sgd      | tagger  |  conll2003   | master  | 171.2         |
2.1           | eager | sgd      | tagger  |  conll2003   | now      | *144*           |
1.15         | decl    | sgd      | tagger  |  conll2003   | master  | 89.5          |
2.1           | decl    | sgd      | tagger  |  conll2003   | master | 93.7           |

PR #514, opened 2020-04-07
fix: correct transition mask when there are `-` in the labels

blester125:
Right now if you have a `-` in your tag label the transition mask function might not create the correct mask, for example if you had two tags `I-X-Y` and `I-X-Z` then when it calls `tag.split("-")[1]` it would return `X` for both tags and the transition would be marked as legal. This is wrong, we should get the types as `X-Y` and `X-Z` and the move between should be marked illegal.

This PR fixes that using the `maxsplit` parameter to `.split` and updates the tests to use fake label names that would trigger this error without the fix.

### DIFF
--- a/layers/eight_mile/utils.py
+++ b/layers/eight_mile/utils.py
@@ -210,15 +210,15 @@ def iob_mask(vocab, start, end, pad=None):
                 if from_.startswith("B-"):
                     # Can't move from a B to a B of another type
                     if to.startswith("B-"):
-                        from_type = from_.split("-")[1]
-                        to_type = to.split("-")[1]
+                        from_type = from_.split("-", maxsplit=1)[1]
+                        to_type = to.split("-", maxsplit=1)[1]
                         if from_type != to_type:
                             mask[vocab[to], vocab[from_]] = small
                 elif from_.startswith("I-"):
                     # Can't move from an I to a B of another type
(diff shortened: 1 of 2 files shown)


layers/eight_mile/utils.py
```diff
@@ -307,13 +307,12 @@ def iobes_mask(vocab, start, end, pad=None):
...
                     elif to.startswith("I-") or to.startswith("E-"):
-                        from_type = from_.split("-")[1]
-                        to_type = to.split("-")[1]
+                        from_type = from_.split("-", maxsplit=1)[1]
+                        to_type = to.split("-", maxsplit=1)[1]
                         if from_type != to_type:
                             mask[vocab[to], vocab[from_]] = small
-                elif from_.startswith(("E-", "I-", "S-", "O")):
+                elif from_.startswith(("E-", "S-", "O")):
                     # Can't move from E to I or E
```
sagnik, line 315: Why did we remove `I-` here? was that a bug before?

blester125: It was superfluous, there is already a case in the if statements that handle `I-` so this being an `elif` meant it was never triggered.

Also if it was ever triggered it would have been a bug because you can't unilaterally say `I -> E` is illegal like the code here does, you need to check the types.

PR #517, opened 2020-04-08
bug fix to handle messy data

wenshuoliu:
Some of the testing data has ` ` after `<EOU>`, causing error. Not sure if it'll happen in training, but we'll better fix it to make sure. I've cut my release to start new job.


wenshuoliu:
Update: Some contexts end with `<EOU> <EOU>`, i.e. the last turn is empty. I think those lines should be thrown away.

PR #519, opened 2020-04-08
WIP: Feature/response selection

wenshuoliu:
This PR intends to add an api-example for response selection job using diffcerent pretrained checkpoint. 

RN it can load a dual-encoder ckpt and do inference on testing data w/o finetuning. 

The current version for CLM model works as the following: it pairs each context in a batch with each of the response and using the CLM model to calculate the conditional joint log probability of the response sequence, normalized by the response length, as the score of the pair. It has problems:
- too slow. it needs more than an hour to run through the testing set with 50k samples. 
- not accurate. so far the 1@100 acc is only around 5%. During debugging, it seems to me the probabilities of each token are reasonable. The problem is that the average of these log probabilities gives too week a signal of the relevance between the response and the context.

### DIFF
--- a/api-examples/response-selection.py
+++ b/api-examples/response-selection.py
@@ -45,6 +45,23 @@ def create_model(model_type, embeddings, d_model=512, d_ff=2048, dropout=0., num
     return model
 
 
+def get_joint_logprob(model: nn.Module, sequences: torch.Tensor, context_length: int):
+    mask = (sequences != 0).to(args.device)
+    inputs = {'x': sequences.to(args.device)}
+    targets = sequences.to(args.device)
+    softmaxes = model.predict(inputs, numpy_to_tensor=False)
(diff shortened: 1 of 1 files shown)


wenshuoliu:
Update: I tried a new score for ranking responses (describe below) and improved 1@100 acc from ~5% to ~15%. 

Here's the ranking process: for a certain context `c_k` and all the responses `r_j (j = 0, ..., bsz-1)` (`r_k` is the correct response), I use `logP(r_j | c_k) - logP(r_j)` as the score for ranking the responses. It adjusts for the probabilities of the response itself. Namely, the correct response may be a rare sequence by it self, but in response selection task we are looking for the relevance between the response and the context, so we should take out the probability of the response.

wenshuoliu:
this change is included in #538, closing it

PR #521, opened 2020-04-09
chore: Added multi-label classification with tf and pytorch

blester125:
This PR adds classes to do multi-label classification. An earlier branch has a lot of conditionals in the model itself for this. This version uses classes to do it. It ends up being pretty clean in pytorch but not quite as clean for tensorflow. It also adds a bunch of multi-label metrics to eight mile

### DIFF
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -115,8 +115,8 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
         probs = self.predict_batch(batch_dict, **kwargs)
         if raw and not dense:
             logger.warning(
-                "Warning: `raw` parameter is deprecated pass `dense=True` to get back values as a single tensor")
-
+                "Warning: `raw` parameter is deprecated pass `dense=True` to get back values as a single tensor"
+            )
             dense = True
         if dense:
             return probs
@@ -163,6 +163,18 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
         if input_dim is None:
             return None
         return Dense(input_dim, output_dim, activation=kwargs.get('output_activation', 'log_softmax'))
+
+
+class MultiLabelMixin(ClassifierModelBase):
+    def init_output(self, input_dim, output_dim, **kwargs):
+        if input_dim is None:
+            return None
+        if kwargs.get('output_activation') is not None:
+            logger.warning("Output activation of `%s` found, this is incompatible with multi-label classification, setting output to Identity")
+        return Dense(input_dim, output_dim, activation=None)
+
+    def create_loss(self):
+        return nn.BCEWithLogitsLoss()
 
 
 @register_model(task='classify', name='default')
@@ -267,3 +279,28 @@ class FineTuneBaselineModel(ClassifierModelBase):
         model.layers = FineTuneModel(len(labels), embeddings, stack_model)
         logger.info(model)
         return model
+
+
+@register_model(task='classify', name='multi-label-conv')
+class MultiLabelConvModel(MultiLabelMixin, ConvModel):
+    pass
+
+
+@register_model(task='classify', name='multi-label-lstm')
+class MultiLabelLSTMModel(MultiLabelMixin, LSTMModel):
+    pass
+
+
+@register_model(task='classify', name='multi-label-nbow')
+class MultiLabelNBowModel(MultiLabelMixin, NBowModel):
+    pass
+
+
+@register_model(task='classify', name='multi-label-nbowmax')
+class MultiLabelNBowMaxModel(MultiLabelMixin, NBowMaxModel):
+    pass
+
+
+@register_model(task='classify', name='multi-label-composite')
+class MultiLabelCompositePoolingModel(MultiLabelMixin, CompositePoolingModel):
+    pass
--- a/baseline/tf/classify/model.py
+++ b/baseline/tf/classify/model.py
@@ -29,6 +29,7 @@ from baseline.tf.tfy import (
 logger = logging.getLogger('baseline')
 
 
+
 class ClassifierModelBase(tf.keras.Model, ClassifierModel):
     """Base for all baseline implementations of token-based classifiers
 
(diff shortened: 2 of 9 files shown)


baseline/tf/classify/training/datasets.py
```diff
@@ -145,3 +145,125 @@ def fit_datasets(model_params, ts, vs, es=None, **kwargs):
         trainer.recover_last_checkpoint()
         trainer.sess.run(test_init_op)
         trainer.test(es, reporting_fns, phase='Test', verbose=verbose)
+
+
+@register_training_func('classify', name="multi-label-datasets")
```
dpressel, line 150: why is this needed?  it doesnt seem to do anything different from the base

baseline/tf/classify/training/utils.py
```diff
@@ -252,3 +253,85 @@ def recover_last_checkpoint(self):
...
+
+            batchsz = len(preds)
+            total_loss += lossv * batchsz
+            total_norm += batchsz
+
+            gold_labels.append(y.astype(np.int32))
+            preds[preds > self.threshold] = 1
+            pred_labels.append(preds.astype(np.int32))
+
+        metrics = ml.get_all_metrics(np.concatenate(gold_labels, axis=0), np.concatenate(pred_labels, axis=0))
```
dpressel, line 334: seems like overkill that we should have to redefine the entire training loop to override the metrics

baseline/pytorch/classify/model.py
```diff
@@ -165,6 +165,18 @@ def init_output(self, input_dim, output_dim, **kwargs):
         return Dense(input_dim, output_dim, activation=kwargs.get('output_activation', 'log_softmax'))
 
 
+class MultiLabelMixin(ClassifierModelBase):
```
dpressel, line 168: this seems reasonable.  OTOH, the way I did this seems like it has minimal changes to overall class structure: https://github.com/dpressel/mead-baseline/compare/feature/v2...feature/multilabel

mead/config/datasets.json
```diff
@@ -143,5 +143,13 @@
     "download": "https://www.dropbox.com/s/lsrdntgu30ftg5l/snips.tar.gz?dl=1",
     "sha1": "c3dd1bff7ad965303065549a05a7f420fc6f3279",
     "label": "snips"
+  },
+  {
+    "label": "conll-multilabel",
```
dpressel, line 148: which dataset is this?

baseline/tf/classify/training/datasets.py
dpressel: in general, I think we should strive to avoid custom `fit_func`s and try and limit them to cases which involve a different sub-system (like eager, distributed, datasets, feed_dict etc.), even if it means some casing within the class.  For example, here you have only overriden the TF 1.x fit for datasets, so does this mean that only this mode supports multi-label?

baseline/tf/classify/model.py
```diff
@@ -360,6 +361,125 @@ def embed(self, **kwargs):
         return EmbeddingsStack(self.embeddings)
 
 
+class MultiLabelMixin(ClassifierModelBase):
```
dpressel, line 364: The only difference between this and the basic classifier is the output layer, isnt there a way to avoid redefining the entire structure?  Maybe instead of using mixins we should assume that its possible for all classifiers to be sigmoid output instead of softmax and have conditions for both paths.  That is more like what I did in my branch, and while I like what you have in the PyTorch, this seems a bit sketchier

baseline/tf/classify/training/utils.py
dpressel: additionally, as we fully support eager mode now (and will probably start recommending it over declarative), there needs to be a full solution for eager and eager-distributed

baseline/tf/classify/model.py
blester125: The main driving change here was that the y placeholder is now `tf.compat.v1.placeholder(tf.float32, [None, nc], name="y")` instead of a `float32` instead of an `int32`. I suppose this class could pass in this as a value for `y` in the kwargs to avoid having to duplicate this function?

baseline/tf/classify/training/datasets.py
blester125: This function needed because of this line

```
model_params['y'] = y
```

In the default implementation it tries to turn the `y` into a one hot vector

```
model_params['y'] = tf.one_hot(tf.reshape(y, [-1]), len(model_params['labels']))
```

But when you are multlihot your `y` is already coming in a a `[B, C]` tensor.

I suppose it would be possible to detect if it was needed based on the shape of `y`? This seems like it would also allow for people to have non sparse labels in the single label situation, like if they were label smoothing or something?

baseline/tf/classify/training/utils.py
blester125: I'll look into added eager stuff, I always forget it about it lol

mead/config/datasets.json
blester125: I took the conll dataset, extracted all the entities for an utterance and the label is the types of the entities, like if an utterance has a 2 locs and a pre the label set would be `PER,LOC`.

It was the easiest dataset I could test all this multilabel stuff on, If you have a better dataset (or we just want to skip having a dataset for it) I can change it

baseline/pytorch/classify/model.py
blester125: My thought was that we have all these hooks in baseline so we should be using them like this when we can. I can look into the more conditional version like your branch.

What was the thought on how we would trigger multilabel? It is just something the reader detects and that percolates up to the model? Or is there something where we can specify the loss and this loss tells us we are multilabel? or just a multilabel flag?

baseline/tf/classify/training/utils.py
blester125: > seems like overkill that we should have to redefine the entire training loop to override the metrics

I agree, what do we think is the best solution? There seems to be a couple ways.

1) Conditional execution based on if the model is multilabel or not.
2) Add hooks that make swapping evaluation functions easier, The training functions have always seemed a little large and hard to customize to me and it seems like this could be a chance to make it more configurable and maybe share things better
3) Overrides like this. (Which we decided was bad lol)

Also one consideration if how this effects future use of things like tf.functions which you mentioned really sped up the taggers.

baseline/pytorch/classify/model.py
dpressel: Yeah, definitely could handle multiple ways.  The reader could attempt to detect from the labels being comma-separated e.g. that its multi-label, and maybe the model could figure it out from the shape of Y as you mentioned.  My concern was that bad interpretation + implicit behavior would be more likely to lead to an upsetting outcome for the user than just explicitly requesting it.  My 2c, is the user knows if they are trying to build a multilabel model, so why not just make them define that explicitly.  But even so its a valid question which block this makes sense in.

baseline/tf/classify/training/datasets.py
blester125: The section of the code where this change is made is analogous to the `make_input` function for the feed_dict code right? I wonder if this part should be customizable via some sort of hook?

baseline/pytorch/classify/model.py
blester125: One consideration of something like a `is_multilabel` flag is it seems to moving away from the model just being a box that is Tensor-In Tensor-Out which is how we have been trying to move towards.

dpressel:
Closing this PR because its requesting a merge to feature/v2 which is going away.  Please reopen on another branch

mead/config/datasets.json
dpressel: This looks like a viable alternative: https://github.com/xiul-msr/e2e_dialog_challenge

Issue #529, opened 2020-04-16
Update TF Transformer layers to support layer norm options


dpressel:
BERT uses the original transformer layer norm locations, and it uses a different value of epsilon than our current LN settings.  I already updated the PyTorch layers to make these options, and validated that we get the same results now as the HuggingFace Transformers given the same checkpoints.  Still need to do this with TensorFlow

PR #537, opened 2020-04-17
Feature/layer norm after option tf

dpressel:
for BERT checkpt load, transform TF needs LN after
    
    BERT is based on the original T2T architecture,
    where the layer norms happen after the transformation.
    
    This makes that optional, and also allows setting eps
    in the layer norm to make sure that BERT checkpoints
    can be loaded.
    
    Also RA attention needs to be supported in decoders,
    for some reason that never got added.


dpressel:
Closes #529

PR #539, opened 2020-04-17
Feature/embeddings reduction

dpressel:
adds a reduction option to `EmbeddingsStack` and makes it available through the mead config with `embeddings_reduction` from the `model` block of the file.

Not as flexible as it could be since we dont currently create a registry for reductions, but it is simpler for now this way.

A `sum` reduction currently requires that the dimensions of each embedding are the same size, which seems reasonable and doesnt hide any internal complexities.  

`concat` reductions remain the default behavior.

A major motivation for this is to make it simpler to use BERT features, which require a position embedding, a word (token) embedding and a token type embedding.  Using a sum reduction in the `EmbeddingsStack` means that we can easily support the composition of a token type and a `pos_word + token` combo (or alternately, if we really want to decompose it: `pos + token + token_type`, though this isnt really necessary since we can always compute a positional embedding internal to the library).

For BERT use-cases where the token type is not set, treating it as a constant bias is faster than doing a `zeros_like` on the input to generate token type indices.  We can just add this bias in as a learnable parameter intead.

One obnoxious difference is that BERT relies on doing a layer norm in the embeddings, b/c it follows the T2T style of placing layer norms at the end of blocks instead of at the beginning.  This measn that the learned positional embeddings with bias need a layer norm block in order to replicate BERT.  This doesnt seem like a big deal, but its a different from the non-bias version that a user might not expect.

One final point -- the reductions are implemented as a Keras layer -- we dont necessarily have to do this, but I didnt see any harm in that.

### DIFF
--- a/addons/embed_bert.py
+++ b/addons/embed_bert.py
@@ -4,9 +4,9 @@ from __future__ import print_function
 
 
 from eight_mile.pytorch.layers import TransformerEncoderStack, subsequent_mask
-from eight_mile.pytorch.embeddings import PyTorchEmbeddings, PositionalLookupTableEmbeddings, LearnedPositionalLookupTableEmbeddings
+from eight_mile.pytorch.embeddings import PyTorchEmbeddings, LearnedPositionalLookupTableEmbeddingsWithBias
 from baseline.embeddings import register_embeddings, create_embeddings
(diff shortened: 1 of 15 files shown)


blester125:
LGTM, I think the reduction as a layer makes sense, in the future a reduction might have a more complex reduction that has parameters so being a layer seems forward thinking

Issue #541, opened 2020-04-20
Simplify baseline models by not using 8mi versions


dpressel:
8 mile provides some simple utility models for each task.  We currently use those, but they have the effect of making the actions performed in each model more opaque than they should be and they have little utility in terms of reusability.  Additionally, when the `call(x)` is employed in Keras layers, this adds another namespace for the sub-models which is a bit counter-intuitive and inconsistent for cases when we need to add our own custom models.

Additionally, the nomenclature between initialization in TF and PyTorch differs unnecessarily.  In both places, the init is roughly the same, but they are named differently, with the PyTorch naming being clearer

PR #542, opened 2020-04-21
Remove 8mi models as internal calls from baseline

dpressel:
make same classify interface between PyTorch/TF

also I updated the comments and added documentation
on the MEAD-Baseline classify architecture to
docs/classify.md

normalize and unify the implementations TF/PyTorch

Tagger ABC follows same flow as Classifier ABCs

update docs on vectorizers

more docs

### DIFF
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -103,7 +103,7 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
 
         return example_dict
 
-    def predict_batch(self, batch_dict, **kwargs):
+    def predict_batch(self, batch_dict: Dict[str, TensorDef], **kwargs) -> TensorDef:
         numpy_to_tensor = bool(kwargs.get('numpy_to_tensor', True))
         examples, perm_idx = self.make_input(batch_dict, perm=True, numpy_to_tensor=numpy_to_tensor)
         with torch.no_grad():
@@ -111,7 +111,7 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
             probs = unsort_batch(probs, perm_idx)
         return probs
 
-    def predict(self, batch_dict, raw=False, dense=False, **kwargs):
+    def predict(self, batch_dict: Dict[str, TensorDef], raw: bool = False, dense: bool = False, **kwargs):
         probs = self.predict_batch(batch_dict, **kwargs)
         if raw and not dense:
             logger.warning(
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -11,6 +11,14 @@ logger = logging.getLogger('baseline')
 
 
 class ClassifierModelBase(nn.Module, ClassifierModel):
+    """Base for all baseline implementations of token-based classifiers
+
+    This class provides a loose skeleton around which the baseline models
+    are built.  It is built on the PyTorch `nn.Module` base, and fulfills the `ClassifierModel` interface.
+    To override this class, the use would typically override the `create_layers` function which will
+    create and attach all sub-layers of this model into the class, and the `forward` function which will
+    give the model some implementation to call on forward.
+    """
 
     def __init__(self):
         super().__init__()
@@ -32,19 +40,14 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
         write_json(self.labels, basename + ".labels")
 
     @classmethod
-    def create(cls, embeddings, labels, **kwargs):
+    def create(cls, embeddings, labels, **kwargs) -> 'ClassifierModelBase':
 
         model = cls()
         model.pdrop = kwargs.get('pdrop', 0.5)
         model.lengths_key = kwargs.get('lengths_key')
-        model.embeddings = embeddings
-        embed_model = model.init_embed(**kwargs)
         model.gpu = not bool(kwargs.get('nogpu', False))
         model.labels = labels
-        pool_model = model.init_pool(embed_model.dsz, **kwargs)
-        stack_model = model.init_stacked(pool_model.output_dim, **kwargs)
-        output_model = model.init_output(stack_model.output_dim if stack_model else pool_model.output_dim, len(labels), **kwargs)
-        model.layers = EmbedPoolStackModel(len(labels), embed_model, pool_model, stack_model, output_model)
+        model.create_layers(embeddings, **kwargs)
         logger.info(model)
         return model
 
@@ -100,9 +103,6 @@ class ClassifierModelBase(nn.Module, ClassifierModel):
 
         return example_dict
 
-    def forward(self, input: Dict[str, torch.Tensor]):
-        return self.layers(input)
-
     def predict_batch(self, batch_dict, **kwargs):
         numpy_to_tensor = bool(kwargs.get('numpy_to_tensor', True))
         examples, perm_idx = self.make_input(batch_dict, perm=True, numpy_to_tensor=numpy_to_tensor)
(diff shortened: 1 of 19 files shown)


dpressel:
Closes #541

docs/tagging.md
```diff
@@ -106,10 +113,58 @@ You can use [`tag-text.py`](../api-examples/tag-text.py) to load a sequence tagg
...
+
+When `mead-train` runs in declarative mode (which is controlled by the `--prefer_eager <bool>` argument), the training block of the MEAD config can contain a key `fit_func: feed_dict` to switch the training loop implementation to the previous default method (this can also be done at run-time by passing `--fit_func feed_dict` as a command-line argument to `mead-train`.
```
blester125, line 170: Should this override be `--x:train.fit_func feed_dict`? I didn't thing we had on off overrides like this anymore

baseline/pytorch/classify/model.py
```diff
@@ -127,68 +127,154 @@ def predict(self, batch_dict, raw=False, dense=False, **kwargs):
             results.append(outcomes)
         return results
 
-    def get_labels(self):
+    def get_labels(self) -> List[str]:
```
blester125, line 130: If we are going a large overhaul of the baseline models do it make sense to remove the fact that the models hold the list of labels? Like instead of passing in the labels the model could just get the number of labels and the labels are basically just another vocab? it seems like that would move towards the tensor in tensor out model idea

baseline/pytorch/tagger/model.py
```diff
@@ -136,83 +129,358 @@ def make_input(self, batch_dict, perm=False, numpy_to_tensor=False):
...
-        return self.layers(input)
-
-    def compute_loss(self, inputs):
-        tags = inputs['y']
-        lengths = inputs['lengths']
-        unaries = self.layers.transduce(inputs)
-        return self.layers.neg_log_loss(unaries, tags, lengths)
-
     def get_vocab(self, vocab_type='word'):
+        """Produce a dictionary for the vocab feature name
```
blester125, line 133: Does anything use this `get_vocab` method anymore? I though all the vocab was coordinated via the services?

baseline/pytorch/tagger/model.py
```diff
@@ -136,83 +129,358 @@ def make_input(self, batch_dict, perm=False, numpy_to_tensor=False):
...
+        else:
+            decoder = TaggerGreedyDecoder(
+                len(self.labels),
+                constraint_mask=constraint_mask,
+                batch_first=True,
+                reduction=kwargs.get('reduction', 'batch')
+            )
+        return decoder
+
+    def init_embed(self, embeddings: Dict[str, TensorDef], **kwargs) -> BaseLayer:
```
blester125, line 274: You have these `init_(embed|encode|decode)` methods duplicated

baseline/pytorch/tagger/model.py
```diff
@@ -136,83 +129,358 @@ def make_input(self, batch_dict, perm=False, numpy_to_tensor=False):
...
+        """
+    def compute_loss(self, inputs):
+        """Define a loss function from the inputs, which includes the gold tag values as `inputs['y']`
+
+        :param inputs:
+        :return:
+        """
+
+
+class TransducerTaggerModel(TaggerModelBase):
```
blester125, line 205: I see that in the docstring there is a mention that encode and transduction mean the same thing but it seems confusing to call the class `Transducer` and the method `init_encode` This should probably match?

docs/tagging.md
dpressel: TBH, I forgot this functionality existed at the time I wrote it.  Im ok with reverting this part, although the alternative is pretty verbose

baseline/pytorch/classify/model.py
dpressel: seems minor, could be revisited in a future PR.  Would rather not decide in this PR

baseline/pytorch/tagger/model.py
dpressel: I think this can go away.  Will double check

baseline/tf/tagger/model.py
```diff
@@ -344,43 +315,132 @@ def create(cls, embeddings, labels, **kwargs):
...
+class TransducerTaggerModel(TaggerModelBase):
 
     def __init__(self):
         super().__init__()
 
-    def encode(self, **kwargs):
+    def create_layers(self, embeddings: Dict[str, TensorDef], **kwargs):
+        self.embeddings = self.init_embed(embeddings, **kwargs)
+        self.encoder = self.init_encode(**kwargs)
+        self.proj_layer = tf.keras.layers.Dense(len(self.labels))
```
blester125, line 347: I know there was some talk of making this projection layer optional for things like the LAN tagger where the encoder output is already the same size as the output but it seems like we can't check for that at build time in TF. Does it make sense to create a `ProjectionLayer` object in `eight_mile` where if the input and output size are the same it does nothing and if they are different it does uses a Dense layer to do the projection?

mead/config/conll-no-crf.json
```diff
@@ -54,7 +54,7 @@
     "dropout": 0.5,
     "rnntype": "blstm",
     "layers": 1,
-    "constrain_decode": true,
+    "constrain_decode": true,
```
blester125, line 57: Extra whitespace added?

baseline/pytorch/tagger/model.py
dpressel: thanks

baseline/pytorch/classify/model.py
blester125: Makes sense

baseline/pytorch/tagger/model.py
dpressel: I dont really like calling it a `Transducer` TBH at all, and the steps are divided into `embed`, `encode` and `decode`.  I dont want to make  it seem like its an encoder-decoder -- its not really, but the transduce just comes from a lack of a better name for the encoder portion thats going on -- calling it encode doesnt seem like it makes sense

baseline/tf/tagger/model.py
dpressel: IMO what makes sense is add an `output_dim` to the BiLSTM encoder in TF like we have in PyTorch.  The proj_layer check could then be a test against the `output_dim` (if they are the same, create, if not dont).  In some case it might still be desirable to explicitly tweak, so another simple change would be to have a `self.init_proj()` step

mead/config/conll-no-crf.json
dpressel: was wondering if you would notice that ;)

docs/tagging.md
dpressel: I guess my gut feeling is it doesnt do any harm.  People who like the long confusing syntax can still use it, but it looks like overriding the `fit_func` will become a lot more common so I dont feel super bad about having modified the driver.

baseline/pytorch/tagger/model.py
dpressel: I guess I dont have a great answer for this.  Whats there now minimally changes the existing interface but I think its an improvement.  If we decide on a better name, happy to make a PR for it, but Id rather not block this PR on that unless we truly hate it

baseline/tf/tagger/model.py
dpressel: I added an `init_proj` layer which you can override if you dont like the linear projection applied.  For instance, you could use a skip conn or you could just use an `PassThru/Identity` in the case that you already have logits coming out of the encoder layer.  I also updated the docs to describe this as "step 3" of the model

baseline/pytorch/classify/model.py
```diff
@@ -127,68 +127,154 @@ def predict(self, batch_dict, raw=False, dense=False, **kwargs):
...
         weight_init = kwargs.get('weight_init', 'uniform')
         rnntype = kwargs.get('rnn_type', kwargs.get('rnntype', 'lstm'))
         if rnntype == 'blstm':
-            return BiLSTMEncoderHidden(dsz, hsz, 1, self.pdrop, unif=unif, batch_first=True, initializer=weight_init)
-        return LSTMEncoderHidden(dsz, hsz, 1, self.pdrop, unif=unif, batch_first=True, initializer=weight_init)
+            return BiLSTMEncoderHidden(input_dim, hsz, 1, self.pdrop, unif=unif, batch_first=True, initializer=weight_init)
+        return LSTMEncoderHidden(input_dim, hsz, 1, self.pdrop, unif=unif, batch_first=True, initializer=weight_init)
 
     def make_input(self, batch_dict):
+        """In PyTorch, we typically set the lengths to sort descending and permute the batch accordingly
```
blester125, line 271: This length permutation stuff has since been added to the base `make_input`. I think we can remove this implementation in the subclass

baseline/pytorch/classify/model.py
```diff
@@ -197,33 +283,90 @@ def make_input(self, batch_dict):
...
+        """
+        hszs = listify(kwargs.get('hsz', []))
+        if not hszs:
+            return PassThru(input_dim)
+        return DenseStack(input_dim, hszs, pdrop_value=self.pdrop)
+
+    def create_layers(self, embeddings: Dict[str, TensorDef], **kwargs):
+        self.embeddings = self.init_embed(embeddings, **kwargs)
+        self.stack_model = self.init_stacked(self.embeddings.output_dim, **kwargs)
+        self.output_layer = Dense(self.stack_model.output_dim, len(self.labels), activation="log_softmax")
```
blester125, line 364: Is there a reason the FineTuneModel doesn't have a hook for overriding the output? It wouldn't be used much but it seems like it would be useful for something like multi-label classification

baseline/tf/classify/model.py
```diff
@@ -76,16 +73,16 @@ def save_md(self, basename):
             if hasattr(embedding, 'save_md'):
                 embedding.save_md('{}-{}-md.json'.format(basename, key))
 
-    def _record_state(self, **kwargs):
+    def _record_state(self, embeddings: Dict[str, TensorDef], **kwargs):
```
blester125, line 76: Is this type hint correct? should it be `Dict[str, BaseLayer]`?

docs/classify.md
```diff
@@ -69,3 +69,49 @@ When reporting the loss reported every nsteps is the total loss averaged over th
...
+
+The only real requirement to provide mead with your own classifier is to extend the ClassifierModelBase from the framework you are using, and to register your model with the `@register_model` annotation to identify a key that uniquely identifies your model.  However, there are some things that make it very easy or succinct to define custom models that we recommend.
+
+#### Some best practices for writing your own classifier
+
+- Use the mead layers (8 mile) API to define your layers
+  - This will minimize any incompatibility and make it easy to switch frameworks later
+  - Determine if your model can sub-class the `EmbedPoolStackClassifer`.  If it follows this idiom, it may be very succinct to define a new classifer by overriding a single function (typically the `init_pool()` function).
+- If you are overriding the `EmbedPoolStackClassifier`, remember that the pooling layer is expecting 2 arguments -- the input tensor itself and a tensor containing the length.  If you want to adapt an existing Layer from Keras/PyTorch or elsewhere that requires a single input Tensor, use the `WithoutLengths(YourLayer)` adapter layer to provide the callee with the input tensor
```
blester125, line 105: The end of this sentence isn't clear to me. something like

> use the `WithoutLengths(YourLayer)` adapter layer so your layer only gets the input tensor and doesn't get the lengths

?

baseline/pytorch/classify/model.py
dpressel: good point, adding that now

baseline/pytorch/tagger/model.py
dpressel: Changed to `AbstractEncoderTaggerModel`

PR #548, opened 2020-04-23
allow sub-layer hooks a name override. closes #547

### LINKED ISSUE #547: TF Tagger and Classifier: allow override of names for Layers
The tagger and classify packages provide idiomatic sub-classes of the `TaggerModelBase` and `ClassifierModelBase` that use `BaseLayer`s to implement a common reusable flow.  The layer implementations provided are also reasonable but they currently overlook the possibility that, in TensorFlow, a user might wish to supply a TensorFlow graph name that is different from the Keras provided one.

The proposed solution for this is to provide a keyword argument (which can be added via mead config) for each known sub-layer.  For instance, in the `init_encode()` in the `RNNTaggerModel`, we would ideally be able to inject the name to the layer as follows:

```

    def init_encode(self, **kwargs):
        """Override the base method to produce an RNN transducer

        :param kwargs: See below

        :Keyword Arguments:
        * *rnntype* (``str``) The type of RNN, defaults to `blstm`
        * *layers* (``int``) The number of layers to stack
        * *hsz* (``int``) The number of hidden units for each layer in the encoder
        * *variational* (``bool``) Variational dropout

        :return: An encoder
        """
        self.vdrop = kwargs.get('variational', False)
        rnntype = kwargs.get('rnntype', 'blstm')
        nlayers = int(kwargs.get('layers', 1))
        hsz = int(kwargs['hsz'])

        Encoder = BiLSTMEncoderSequence if rnntype == 'blstm' else LSTMEncoderSequence
        return Encoder(None, hsz, nlayers, self.pdrop_value, self.vdrop, name=kwargs.get("encode_name"))

```

dpressel:
In some cases we would like to be able to override the Keras layer name behavior for sub-modules.
The sub-class implementations of `TaggerModelBase` and `ClassifierModelBase` give us `init_*()` methods that create the sub-layer objects for us, but up until this change there wasnt any way to
override the name scope without actually reimplementing the method which is not convenient.

The solution proposed here is for each method, provide an optional keyword argument that can be overriden either by mead configuration (from the model block) or by overriding the `create_layers` method to inject this name if known in code.

*Classifier Layer Name Overrides*

- `init_embed` -> `embeddings_name` (injects name into `EmbeddingsStack`)
- `init_stacked` -> `stacked_name` (injects name into `DenseStack`)
- `init_output` -> `output_name` (injects name into `tf.keras.layers.Dense`)
- `init_pool` -> `pool_name` (injects name into whatever the sub-class pool Keras layer is defined as)

*Tagger Layer Name Overrides*

- `init_embed` -> `embeddings_name` (injects name into `EmbeddingsStack`)
- `init_encode` -> `encode_name`  (injects name into whatever the sub-class pool Keras layer is defined as)
- `init_proj` -> `proj_name` (injects name into `tf.keras.layers.Dense`)
- `init_decode` -> `decode_name` (injects name into `GreedyTaggerDecoder` or `CRF` in base impl.)

Also changes the Classifier `init_embed` in both PyTorch and TF to accept an optional `embeddings_dropout` which defaults to 0 that can be passed in.  The tagger currently uses the default pdrop for this, although perhaps that should change as well to `embeddings_dropout` 2with maybe a default of 0.5 in that case

### DIFF
--- a/baseline/pytorch/classify/model.py
+++ b/baseline/pytorch/classify/model.py
@@ -165,13 +165,19 @@ class EmbedPoolStackClassifier(ClassifierModelBase):
 
     def init_embed(self, embeddings: Dict[str, TensorDef], **kwargs) -> BaseLayer:
         """This method creates the "embedding" layer of the inputs, with an optional reduction
+
+        :param embeddings: A dictionary of embeddings
+
         :Keyword Arguments: See below
         * *embeddings_reduction* (defaults to `concat`) An operator to perform on a stack of embeddings
+        * *embeddings_dropout = float(kwargs.get('embeddings_dropout', 0.0))
 
         :return: The output of the embedding stack followed by its reduction.  This will typically be an output
           with an additional dimension which is the hidden representation of the input
         """
-        return EmbeddingsStack(embeddings, reduction=kwargs.get('embeddings_reduction', 'concat'))
+        reduction = kwargs.get('embeddings_reduction', 'concat')
+        embeddings_dropout = float(kwargs.get('embeddings_dropout', 0.0))
+        return EmbeddingsStack(embeddings, embeddings_dropout, reduction=reduction)
 
     def init_pool(self, input_dim: int, **kwargs) -> BaseLayer:
         """Produce a pooling operation that will be used in the model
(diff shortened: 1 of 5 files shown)


dpressel:
To use this, you can update the `model` block of the config and it will update the name scopes accordingly:

```
 "model": {
        "model_type": "default",
        "hsz": 200,
        "dropout": 0.5,
        "rnntype": "blstm",
        "layers": 1,
        "encode_name": "encoder",
        "decode_name": "decoder",
        "embeddings_name": "es",
        "output_name": "output",
        "crf": true 
    },
```
trainable_vars

PR #551, opened 2020-04-25
chore: make eager debugging easier to use

blester125:
With the addition of `tf.function`s and the like in mead (which was def
worth it due to speed ups) it makes eager debugging harder to use. When
something is decorated with a `tf.function` the code will be compiled
into a graph by tensorflow. This allows for faster execution but makes
step-by-step debugging harder (They even issue an error when they detect
a call to `pdb.set_trace()` in your code).

This PR introduces a new function to `eight_mile.tf.layers` that will
disable the compilation of `tf.function`s. This is trigger-able with an
env variable `MEAD_TF_EAGER_DEGUB`. Setting this to a truthy value (as
parsed by `str2bool`) will make debugging easier by keeping the
functions in eager mode which lets you inspect the values directly in
your debugger

### DIFF
--- a/layers/eight_mile/tf/layers.py
+++ b/layers/eight_mile/tf/layers.py
@@ -27,6 +27,13 @@ def set_tf_eager_mode(prefer_eager: bool = False):
         LOGGER.info('User requesting eager disabled on 2.x')
         tf.compat.v1.disable_eager_execution()
 
+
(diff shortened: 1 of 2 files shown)


dpressel:
Good idea, was thinking of doing this a while back but lost track of it.

Issue #553, opened 2020-04-25
Update LM API to be consistent with the others


dpressel:
The LM has been in a pseudo-1.x state for a long time.  In TF, the layers API was partially implemented and in the PyTorch nothing really had changed.  Update so that we have

- `LanguageModelBase` has a `create_layers()` abstract method
- An `AbstractDecoderModel` provides the embeddings, embedding projection, (optionally weight-tied) output, and an abstract decoder
- RNN and Transformer flavors should be updated to inherit the abstract decoder and to override the minimal amount of code

PR #554, opened 2020-04-26
Changes the API for LMs to be closer to tagger

dpressel:
clean up the LM

* generic name (generator)
* refactor TLM serialization to check old and new attr
* fix broken TLM code
* change names between PyTorch and TF to be same
* update w/ MLM for TF


dpressel:
Closes #553

PR #559, opened 2020-04-28
WIP: chore: added posterior decoding and updating the tagger decoder API, add tests

blester125:
This PR updates the API for the tagger decoders. There is still some more work to do but I wanted to get comments on it earlier rather than later.

 * It creates a separation between taggers that do structured prediction vs locally normalized models.
 * It adds to the decoder API to:
    * get normalization factor at each timestep
    * get the normalization factor going backwards
    * Get the posterior distribution from the taggers
    * Decode based on the posterior
 * It adds a lot of tests
    * Adds tf2 tests for the tagger decoders
    * Adds more batch stability tests to the pytorch decoder
    * Add tests that compare the computed values between the tf/pyt version and explicit python code
 * It also sets the `EOS` tag at the end of the unary/tags for tensorflow so that we don't output illegal tags at the end of the sequence (`B-X` at the end, etc.)

This doesn't touch the main path though the code (the partition function and the viterbi decoder) so I don't think is should a big burden to merge.

There are still a few things to do before merging.

- [ ] Add the partition functions to the tensorflow `ConstrainedGreedyTaggerDecoder`
- [ ] Add pytorch tests for the constrained decoder taggers
- [ ] Add test for tf1
- [ ] Add tensorflow tests for the constrained decoder taggers
- [ ] Add pytorch tests for normal greedy decoder (score sentence, etc)
- [ ] Add tensorflow tests for normal greedy decoder (score sentence, etc)

### DIFF
--- a/baseline/pytorch/tagger/model.py
+++ b/baseline/pytorch/tagger/model.py
@@ -50,6 +50,7 @@ class TaggerModelBase(nn.Module, TaggerModel):
         model.gpu = False if device == 'cpu' else model.gpu
         return model
 
+
     def drop_inputs(self, key, x):
         """Do dropout on inputs, using the dropout value (or none if not set)
         This works by applying a dropout mask with the probability given by a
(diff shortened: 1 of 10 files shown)


dpressel:
Closing as feature/v2 branch is getting removed

Issue #546, opened 2020-04-22
Simplify Embeddings


dpressel:
Right now, we have 2 layers of Embedding objects -- the layers themselves and the model.

In PyTorch, the model just does an empty sub-class on most Embeddings but then `register_embeddings` on the model.  In TF the models currently

1. own an underlying layer and exercise their forward `call(x)`
2. provide additional state recording functionality
3. provide shallow graph copy support since the embeddings are created in another graph
4. provide placeholder creation functionality, and also `register_embeddings` on the model.
5. provide an underlying `.encode(x)` that is called by `call(x)` in the base

We would like to eliminate most of this and push down other parts into layers.

1. at a minimum, we'd like a similar inheritance to PyTorch b/c its less confusing about where to derive than owning a sub-object.  Ideally, we could eliminate the baseline Model embeddings altogether and `register_embeddings` directly in the base Embeddings layers
2. the functionality for state recording on Embeddings models is critical to rehydrate them properly.  We could either try and push this down into layers, or do some mixing in to get the functionality later
3. the reference seems not to matter much anymore, perhaps only mattered for multi-GPU (especially in 1.x).  Distributed mode is supported currently only with eager, and this seems to be a non-issue (need to verify)
4. The placeholder creation functionality seems like it could be externalized.  Instead of having a classmethod producing the placeholder on demand (and also we do this inside `.encode(x)` if x is `None`), it seems that we could have the embeddings just declare their required number of dimensions and possibly their `dtype` if it needs an override.  An external `baseline.tf.placeholder_for(embedding)` seems like it could solve the coupling issue,  and we would modify the Embeddings to not allow a null argument for `x`.  This would change the API slightly but not really have much negative impact that I can think of
5. Eliminate `encode(x)` as its use is legacy

dpressel:
The #560 PR is a rather simple as it mainly refactors *Models to inherit *Embeddings

Covers 1. with inheritance, 2. as a mixin and 3. provisionally (though needs more testing to verify)

Issue #528, opened 2020-04-16
Loading BERT checkpoints into TLM from TensorFlow


dpressel:
We have a solution for loading BERT checkpoints into PyTorch 8 mile layers directly from huggingface checkpoints by modifying the keys in the pytorch state dict.  What would be better is to use this code to then re-write the checkpoint in TLM NPZ format so it can be loaded into TensorFlow 8 mile layers as well.  The NPZ files can be hosted on dropbox and has the advantage that the same checkpoint can be used in pytorch and TF

dpressel:
This PR makes TF support the same options as PyTorch including layer norm epsilon and T2T-style LN (after) https://github.com/dpressel/mead-baseline/pull/537

dpressel:
Waiting on checkpoint generation for verification

PR #560, opened 2020-04-28
embeddings refactor with NPZ bert loads

dpressel:
This change provides a converter from huggingface transformers to NPZ format which can be used by Transformer LMs and fine-tuning directly using the MEAD Transformer layers.  Once the change is accepted, these converted checkpoints will be uploaded to make it easy to use.

Also this change eliminates the need for a BERT-specific addon -- now BERT is just a special case of `embed_tlm_*` where the layer norms happen after the layers and the embeddings include an LN reduction.  

As part of the change, it was easiest to refactor the embeddings to follow the same approach as the
pytorch.  As part of that, some issues were identified and fixed with the naming.

There is now a mixin to supply the additonal required interfaces for TF.  The models now subclass the original layers and the mixin to get state recording.

As part of the change, calls to shallow copy the embeddings sub-graphs were removed -- it doesnt appear that these are required at the moment, although we will run multi-GPU testing to confirm.

One note: The `embed_tlm` doesnt (cannot?) support custom token type embeddings at the momemt.  Its possible to use these but I think it probably requires extending the model itself, not the embedding.  In the future we probably should support multiple features per embedding


dpressel:
Closes #546 and #528

baseline/tf/embeddings.py
```diff
@@ -120,116 +82,86 @@ def get_config(self):
...
         return tf.compat.v1.placeholder(tf.int32, [None, None, None], name=name)
 
 
 @register_embeddings(name='char-transformer')
-class CharTransformerModel(TensorFlowEmbeddingsModel):
-    def __init__(self, name=None, **kwargs):
-        super().__init__(name, **kwargs)
-        self.embedding_layer = CharTransformerEmbeddings(name=self._name, **kwargs)
+class CharTransformerModel(CharTransformerEmbeddings, TensorFlowEmbeddingsMixin):
+    pass
```
blester125, line 102: Why does this embedding not need a create placeholder?

baseline/tf/embeddings.py
```diff
@@ -26,49 +24,18 @@ def detached_ref(self):
...
         self.x = x
-        return self.embedding_layer(x)
-
-    def call(self, x):
-        return self.encode(x)
 
-    def get_feed_dict(self):
-        """Return a feed dict that is needed to initialize this embeddings."""
-        return self.embedding_layer.get_feed_dict()
+        return super().encode(x)
```
blester125, line 38: I thought part of the plan of simplifying the embedding was to remove the `encode` and replace it with `call`?

layers/eight_mile/tf/layers.py
```diff
@@ -1792,7 +1795,7 @@ def items(self):
     def __getitem__(self, item: str):
         return self.embeddings[item]
 
-    def call(self, inputs):
+    def call(self, inputs: Dict[str, tf.Tensor]) -> tf.Tensor:
```
blester125, line 1798: I thought we were using that `TensorDef` for type hints about tensors now?

PR #566, opened 2020-04-29
chore: mead can order hours and minutes if need be

blester125:
This PR lets the mead dataset finder parse datasets that have hour and minute time stamps too. When there is a date with no hour/minutes then the datetime for those defaults to 00 and 00 which means that a new one with the hours/minutes would be returned because it is later in time which seems like the right behavior

### DIFF
--- a/mead/utils.py
+++ b/mead/utils.py
@@ -17,7 +17,18 @@ export = exporter(__all__)
 logger = logging.getLogger('mead')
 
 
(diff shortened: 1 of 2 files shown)


mead/utils.py
```diff
@@ -16,6 +17,9 @@
 logger = logging.getLogger('mead')
 
 
+KNOWN_FMTS = ['%Y%m%d', '%Y-%m-%d', '%Y/%m/%d', '%Y', '%Y%m', '%Y-%m', '%Y/%m', '%Y%m%d_%H%M', '%Y-%m-%d_%H-%M', '%Y/%m/%d_%M/%H']
```
dpressel, line 20: not very descriptive: `KNOWN_DATE_FORMATS`?

blester125: Fixed

PR #572, opened 2020-05-01
Fix/convert model

wenshuoliu:
Changes based on our discussion:
- Add skip connection and layer norm as options to DenseStack; activation can be a list. For both pytorch and tf. 
- Move `create_model` to `transformer_utils.py`. Keep the name since it also creates seq2seq model.
- Put `SingleHeadReduction` in eight mile layers, considering it's a fancy pooling and might be useful somewhere else. 
- `response-selection.py` only works for dual-encoder model right now for simplicity. 

Changes slightly different from our discussion:
- Keep `TwoHeadConcat` in `transformer_utils.py` since it's very specific to dual-encoder model. 
- Add a customized module `ConveRTFFN` to still implement the convert graph. Looking at it again, I think the convert graph does make some sense, though not compatible with the DenseStack we want. E.g. the reason of layer norms at both beginning and end is because the `final` layer doesn't have either activation or layer norm. And the "projecting skip-connection" adds two unnormalized tensor before going into the final layer norm. I don't want to make much changes to the model structure since I've been experimenting on it.

### DIFF
--- a/api-examples/transformer_utils.py
+++ b/api-examples/transformer_utils.py
@@ -21,6 +21,7 @@ def find_latest_checkpoint(checkpoint_dir: str, wildcard="checkpoint") -> str:
             step_num = this_step_num
     return checkpoint
 
+
 class TripletLoss(nn.Module):
     """Provide a Triplet Loss using the reversed batch for negatives"""
     def __init__(self, model):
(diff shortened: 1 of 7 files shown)


api-examples/transformer_utils.py
```diff
@@ -407,3 +458,29 @@ def forward(self, inputs, targets):
                 pred = self.model(in_)
                 return self._loss(pred, targets)
         return LossFn(self, loss)
+
+
+def create_model(embeddings, d_model, d_ff, dropout, num_heads, num_layers, model_type, rpr_k, d_k, reduction_d_k,
```
dpressel, line 463: create_paired_model?

baseline/pytorch/embeddings.py
```diff
@@ -208,11 +208,15 @@ def _identity(x):
     return x
 
 
-def _mean_pool(_, embeddings):
+def _mean_pool(inputs, embeddings):
```
dpressel, line 211: tf impl?

baseline/pytorch/embeddings.py
```diff
@@ -208,11 +208,15 @@ def _identity(x):
...
+    mask = (inputs != 0)
+    embeddings = embeddings * mask.unsqueeze(-1)
     return torch.mean(embeddings, 1, False)
 
 
-def _max_pool(_, embeddings):
+def _max_pool(inputs, embeddings):
```
dpressel, line 217: dont really need this I think

baseline/pytorch/embeddings.py
```diff
@@ -233,8 +237,10 @@ def __init__(self, name, **kwargs):
             self.pooling_op = self._cls_pool
 
     def _sqrt_length_pool(self, inputs, embeddings):
```
dpressel, line 239: tf?

api-examples/transformer_utils.py
wenshuoliu: this also create seq2seq model, so i'd like to keep it

dpressel: seq2seq is still paired.  the problem is the current name is not specific

dpressel:
Im going to close this for now to cut another RC, but please remove the sqrt pool from the pytorch model and rename `create_model` to `create_paired_model` when you get a chance.

PR #575, opened 2020-05-01
Fix/bert ft onnx

dpressel:
We were doing a boolean array indexing to find the `[CLS]` index which isnt supported by ONNX and causes and invalid model to be generated.

Also, we were using fake random number data to generate the tensors which means that tokens like `[CLS]` where unlikely to be generated.  The fix for this is to stop generating numbers and use fake sentence and run them through the vectorizers


dpressel:
I tested this with an LSTM classifier, CNN-BiLSTM-CRF and a BERT fine-tuned model.  All seem to work with this configuration

PR #576, opened 2020-05-03
fixes broken reload of modules on mead-hub.

dpressel:
When we go to reload modules that were downloaded from mead-hub, we encounter the issues that they are not in the `sys.path`.  Since all addons from mead-hub go in the `~/.bl-data/addons` directory, we can solve this problem just by prepending to the path (we have to prepend because the CWD is usual on the front, which can cause issues if a model directory is in the path).

We have code currently in `import_user_module()` that brings in the installed `addons` path, but this is no longer the primary location where modules live.

This PR creates a function `put_addons_in_path()` and changes the API examples to make this call.
It also wouldve been possible to place this call right in `import_user_module` alongside the portion whee we add `addons`, but that feels like a bad side effect (in fact, I think we should change the current behavior of always adding this path).

This doesnt completely answer the question of what to do for inference on another machine, but presumably, the client code could do an upfront download of hub to addons to be safe.

We discussed other more complex ways of handling this via hub like keeping an index (although we already have that in `data-cache.json` and unless we store the source location in the model state, we wouldnt be able to recover it.

TBD how to handle where the training machine is different from
the inference machine

### DIFF
--- a/api-examples/classify-text-onnx.py
+++ b/api-examples/classify-text-onnx.py
@@ -7,9 +7,10 @@ from baseline.utils import str2bool, read_json
 import numpy as np
 from collections import defaultdict
 import onnxruntime as ort
-from baseline.utils import unzip_files, find_model_basename, load_vectorizers, load_vocabs
+from baseline.utils import unzip_files, find_model_basename, load_vectorizers, load_vocabs, put_addons_in_path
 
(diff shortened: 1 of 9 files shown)


dpressel:
Using #577 instead

Issue #581, opened 2020-05-05
Test/update mead configs


dpressel:
- [x] ag-news-lstm
- [x] ag-news-single-lstm
- [x] ag-news
- [x] conll-bert
- [x] conll-cnn
- [x] ~~conll-elmo-hub~~
- [x] conll-elmo
- [x] conll-no-crf
- [x] conll-single-tf
- [x] conll-single
- [x] conll-stacked
- [x] conll-transformer
- [x] conll
- [x] dbpedia
- [x] iwslt15-en-vi
- [x] iwslt15-en-vi-transformer
- [x] mrda
- [x] mrda-hier
- [x] ontonotes
- [x] ~~ptb-convchar~~
- [x] ptb-kim-adam
- [x] ptb-med
- [x] ptb-small-tf
- [x] ptb-transformer-chars (broken!)
- [x] ptb-transformer
- [x] reverse-transformer
- [x] reverse
- [x] snips-intent
- [x] snips
- [x] sst2-bert-base-uncased
- [x] sst2-bert-finetune-pytorch (rename?)
- [x] sst2-bert-hub
- [x] sst2-bert-pytorch
- [x] sst2-demolib
- [x] sst2-distributed
- [x] sst2-elmo-eh
- [x] ~~sst2-elmo-hub~~
- [x] sst2-elmo-large (requires a PR to hub, but working locally)
- [x] ~~sst2-elmo~~
- [x] sst2-fine-tune-bert (rename?)
- [x] sst2-lstm-840b
- [x] sst2-lstm-char
- [x] ~~sst2-lstm-nolengths~~
- [x] sst2-lstm
- [x] sst2-nogpu-pyt
- [x] twpos-nogpu-pyt
- [x] sst2
- [x] trec-bert-base-uncased
- [x] trec-bert-finetune-pytorch-eh
- [x] trec-bert-finetune-tf-eh
- [x] trec-cnn
- [x] trec-demolib
- [x] trec-lstm
- [x] twpos-blstm-lan
- [x] twpos-stacked
- [x] twpos
- [x] wnut-bert
- [x] wnut-elmo
- [x] wnut-no-crf
- [x] ~~wnut-pyt~~
- [x] wnut-single-no-crf
- [x] wnut
- [x] ~~word-char-sst2-pyt~~
- [x] ~~word-char-sst2-tf~~

dpressel:
[...]

PR #587, opened 2020-05-05
fix: sst2-bert-fine-tune config

sagnik:


sagnik:
related to https://github.com/dpressel/mead-baseline/issues/581

dpressel:
can we rename this to `sst2-bert-official-base-uncased`?

Issue #432, opened 2020-01-24
bundle tensorflow serving protobufs better


blester125:
Right now the way we use the tf serving protobufs is the because we pip install with -e the python interpreter can look for modules in `$BASELINE/python` and can find the tensorflow serving directory. When we pip install from pypi we don't want to bundle that because it would overwrite their tf serving install if they had it. This should be moved inside of the baseline source tree and loaded in a better way if we want to continue bundling instead of just having tf serving as a dep if they install with the remote option.

PR #588, opened 2020-05-05
chore: move where tensorflow_serving code lives

blester125:
So we have a dir called `tensorflow_serving` that have code
autogenerated from the tensorflow serving proto buffs. These protobuffs
were generated by Barta in Nov 2018.

This dir is peer level to the `baseline` directory so when we build the
pip package the `tensorflow_serving` becomes a package that will get
installed along with baseline. For example the act of running `pip
install mead-baseline` will create a `baseline` dir and a
`tensorflow_serving` dir in your `site-packages`. The problem with this
is that you can also install `tensorflow-serving-api` to install a
tensorflow_serving package. If you already have this package installed
(you need it for other things etc) then our install will overwrite some
of the files you installed before. Going the other way if you have out
code installed tensorflow serving will overwrite it and you will be
making calls with slightly different code.

This PR moves the tensorflow serving code to live under the baseline
name so that when we install it it is protected from clashing with the
normal tf serving api. The code was also updated to deal with this.

This PR is based on the assumption that there is something special about
these versions of the api that we need. We could also look into just
using the versions installed with `pip install tensorflow-serving-api`
we already have a dependency on `grpc` that needs to be installed when
doing grpc calls (this is normally installed as part of the tensorflow
install so it was an explicit dep until I added it) so it might not be
bad to just use there code. It would remove this weird bundling we would
need to test that their code isn't too different/it works. I assume they
would use some sort of versioning to keep these client apis in sync with
the tf serving servers they release. With the way it is right now the
baseline grpc code would break if tf serving ever changed it's api

closes https://github.com/dpressel/mead-baseline/issues/432


dpressel:
LGTM

PR #591, opened 2020-05-05
if we reload loaded mods we can get an error

dpressel:
This change allows us to specify the same addon multiple times without having to reload it.  If we dont do this we would get an error the second time.  Its also inefficient to do an HTTP GET for the same module 2x.

Typically occurs if the vectorizer and the embeddings both refer to the same addon (like elmo).


dpressel:
This would be a little less scary if it didnt return `None`.  However, no downstream code seems to rely on this for training.  I suggest that we consider changing the `MEAD_HUB_MODULES` to a dict and then serialize its keys for the next version.

PR #603, opened 2020-05-07
fixes #602 and parallel conv features

### LINKED ISSUE #602: Add support back for LMs which generate a different feature than the input ones
We used to support (until very late in the 2.x branch), Kim et al 2015 (also ELMo) models as LMs, where we used one set of features for the src embeddings and one for the destination embeddings.  In the case of the Kim model specifically, we are tasked with generating the words, but the input features are characters, which makes the model very efficient for fine-tuning since it has many fewer parameters and the final softmax LM head is removed.

This was broken during the refactoring.  Its quite straightforward to add back the support -- I think we just need a custom `EmbeddingsStack` that can mask the input.

dpressel:
- adds a `vsz` field on LMs so they dont need to track
self.embeddings[tgt_key]
- adds an `embeddings` arg to `init_output` so
weight-tying still works
- fixes the TensorFlow impl of CharConvEmbeddings when there
are multiple filters specified.  This wasnt hit during
tagging because the default filters are length=3, not variable.
During testing of the Kim LM which uses multiple parallel
features, this issue was uncovered

### DIFF
--- a/baseline/tf/lm/training/eager.py
+++ b/baseline/tf/lm/training/eager.py
@@ -13,7 +13,7 @@ from baseline.tf.lm.training.utils import to_tensors, SHUF_BUF_SZ, NUM_PREFETCH
 
 def loss_with_state(model, h, x, y):
     logits, h_out = model(x, h)
-    vsz = model.embeddings[model.tgt_key].get_vsz()
+    vsz = model.vsz
     targets = tf.reshape(y, [-1])
     bt_x_v = tf.nn.log_softmax(tf.reshape(logits, [-1, vsz]), axis=-1)
     one_hots = tf.one_hot(targets, vsz)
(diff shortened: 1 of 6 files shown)


baseline/tf/lm/training/eager.py
```diff
@@ -13,7 +13,7 @@
 
 def loss_with_state(model, h, x, y):
     logits, h_out = model(x, h)
-    vsz = model.embeddings[model.tgt_key].get_vsz()
+    vsz = model.vsz
```
blester125, line 16: Does it make sense to call this `vsz`? In other places we call it things like nc or num_tags. Should we standardize one something like `output_dim` for these models like we have on the layers?

dpressel: I dothink it would probably make unify the fields, however, The vocab size value (`vsz`) Is used pretty consistently elsewhere (in the embeddings), and I'd hesitate to unify all and risk breakage before the next major release.  We could potentially define `def output_dim()` properties on the models to prepare them for this, which could, in turn, delegate to the current names, which would make it easier to phase them in, but I do think that this is probably outside of the scope of the current PR

PR #612, opened 2020-05-11
support non-directory/non-zip checkpoint names

dpressel:
sometimes there might be multiple pid names in
the same file. this happens when multiple runs are
done and then the user needs to load only one.

by checking the job status or even just the time
of each checkpoint written out, the user can easily
determine the PID-specific job name, but its not
currently loadable.

This change allows the user to give a longer expression
as the checkpoint name and resolves that to the full
checkpoint name, and passes this along to the suffix

### DIFF
--- a/baseline/pytorch/torchy.py
+++ b/baseline/pytorch/torchy.py
@@ -85,29 +85,3 @@ def append2seq(seq, modules):
 
     for i, module in enumerate(modules):
         seq.add_module('%s-%d' % (str(module).replace('.', 'dot'), i), module)
-
-
-def tensor_max(tensor):
-    return tensor.max()
-
-
-def tensor_shape(tensor):
-    return tensor.size()
-
-
-def tensor_reverse_2nd(tensor):
-    idx = torch.LongTensor([i for i in range(tensor.size(1)-1, -1, -1)])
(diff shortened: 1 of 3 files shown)


baseline/utils.py
```diff
@@ -324,7 +324,7 @@ def save_vocabs(basedir, embeds_or_vocabs, name='vocabs'):
 
 
 @export
-def load_vocabs(directory):
+def load_vocabs(directory: str, suffix: Optional[str] = None):
     vocab_fnames = find_files_with_prefix(directory, 'vocabs')
```
sagnik, line 328: wouldnt you pass in `suffix` here?

baseline/pytorch/torchy.py
```diff
@@ -85,29 +85,3 @@ def append2seq(seq, modules):
 
     for i, module in enumerate(modules):
         seq.add_module('%s-%d' % (str(module).replace('.', 'dot'), i), module)
-
```
sagnik, line 88: are these changes intended?

dpressel: yeah, just culling some unused code

baseline/utils.py
dpressel: oops you are right. updated

PR #617, opened 2020-05-15
Fix/api ex class

dpressel:
fix and clean up examples using `8 mile` directly to do classification.
sped up the TF version with `tf.function` as well


dpressel:
broken build is due to some issue between TF 2.1 and TFA

PR #618, opened 2020-05-17
WIP: feature/simplify tf seq2seq

dpressel:
There are some major challenges with continuing to support both declarative mode and eager mode with seq2seq.  The official TF way for about a year has been the eager way, which is basically exactly
how we have it in PyTorch.  This PR brings the 2 implementations together in feel and fixes broken eager mode loading, but it also eliminates support for v1 declarative mode so its a big step and Im not 100% sure we want to do this right now.

If we dont, its possible to do something in-between that applies these fixes but unbreaks declarative mode, but the code will probably get a bit messy.


dpressel:
This is pretty stale at this point.  Might revisit later

PR #628, opened 2020-05-20
add unfolded implementation of RA masking

wenshuoliu:


layers/eight_mile/pytorch/layers.py
```diff
@@ -2839,7 +2905,10 @@ def forward(self, qkvm: Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Te
         key = self.w_K(key).view(batchsz, -1, self.h, self.d_k).transpose(1, 2)
         value = self.w_V(value).view(batchsz, -1, self.h, self.d_k).transpose(1, 2)
 
-        rpr_key, rpr_value = self.make_rpr(seq_len, query.device)
+        if self.ra_masking:
+            rpr_key, rpr_value = self.get_rpr(query.device)
```
dpressel, line 2909: the naming here is confusing.  should be more descriptive, when we are using a limited horizon mask, we have `get_rpr` and when we use Shaw style, we have `make_rpr`.  How about something `make_windowed_rpr()`?

api-examples/pretrain-paired.py
```diff
@@ -55,6 +55,7 @@ def train():
     parser.add_argument("--stacking_layers", type=int, nargs='+', default=[1024, 1024, 1024],
                         help="Hidden sizes of the dense stack (ff2 from the convert paper)")
     parser.add_argument("--ff_pdrop", type=float, default=0.1, help="Dropout in the dense stack")
+    parser.add_argument("--ra_masking", type=str2bool, default=False, help="whether prevent attention beyond rpr_k")
```
dpressel, line 58: this term doesnt make a lot of sense to me, what about `windowed_ra` or `local_ra` or something that is more descriptive of the purpose

layers/eight_mile/pytorch/layers.py
```diff
@@ -2669,6 +2669,61 @@ def _attention(
...
+    def _attention(
+            self, query: torch.Tensor, key: torch.Tensor, rpr_key: torch.Tensor, mask: Optional[torch.Tensor] = None
+    ) -> torch.Tensor:
+        """Implementation of attention considering RA masking: using torch.Tensor.unfold to create an extra dimension
+        representing the sliding window. Then when applying matmul, Q, K, V share the same T dimension.
+
+        :param query: a query for alignment. Can come from self in case of self-attn or decoder in case of E/D
+        :param key: a set of keys from encoder or self
+        :param mask: masking (for destination) to prevent seeing what we shouldnt
+        :param rpr_key: tensor of the rpr_key embeddings [window_sz, d_k]
```
dpressel, line 2698: comments refer to both `window_sz` and `W` but it looks like they are the same.  Lets pick on and update the comments to be consistent

dpressel:
- The TensorFlow version needs to be wired up to support windowed RA in the Encoders
- Both frameworks appear to need updates to support in the Decoders
- Putting the `windowed_ra` in the middle of the parameters of the existing Encoders could create issues for backwards compat if named params are not being used, we should move this to the back.

dpressel:
The TensorFlow interface to unfolding is a bit cumbersome, and it would be nice if the frameworks had a common function between them to execute the unfolds here.  Im thinking we can pull up a utility (function) with same name and signatures to use in each `layers` file to just call so the logic isnt all smack dab in the middle of the attention impl.

PR #632, opened 2020-05-22
make a copy of label across devices

wenshuoliu:


wenshuoliu:
> what about here?
> 
> https://github.com/dpressel/mead-baseline/blob/7a543e93caeb99457ec9f257605bcb46f067086d/api-examples/pretrain-discrim.py#L429

i think you meant L411. Please see the update.

PR #634, opened 2020-05-26
Feature/pretrain refactor

wenshuoliu:
- add lr scheduler options
- save ckpts before finishing an epoch for pretrain-transformer-lm.py

### DIFF
--- a/api-examples/transformer_utils.py
+++ b/api-examples/transformer_utils.py
@@ -645,22 +645,6 @@ def create_model(embeddings, d_model, d_ff, dropout, num_heads, num_layers, mode
     return model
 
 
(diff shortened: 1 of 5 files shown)


api-examples/transformer_utils.py
```diff
@@ -642,3 +643,37 @@ def create_model(embeddings, d_model, d_ff, dropout, num_heads, num_layers, mode
 
     logger.info(model)
     return model
+
+
+def save_checkpoint(model: torch.nn.Module, model_base: str, count: int, logger, tick_type: str = 'epoch'):
```
dpressel, line 648: This was already moved into 8mi

PR #637, opened 2020-05-28
Feature/tlm simplifications

dpressel:
* add a helper in utils for getting number of workers
* add an example that trains Transformer LMs (like RoBERTa) with distributed eager mode
* move `Average` util outside of PyTorch subsystem


dpressel:
A few items on the TODO list here:

- [ ] Support TPU distribution
- [ ] Support LRS as in PyTorch
- [ ] Export NPZ checkpoints if possible so we can start again in PyTorch
- [ ] Fix issue: `Error occurred when finalizing GeneratorDataset iterator: Cancelled: Operation was cancelled`

PR #648, opened 2020-06-07
adds shuffling to distributed training

dpressel:
the distributed worker readers werent shuffling per thread in PyTorch. 
Add that, allowing shuffling of file set and shuffling for each file and changed allocation
of reads to distribute more evently.
Also add LR logging on update steps

### DIFF
--- a/api-examples/transformer_utils.py
+++ b/api-examples/transformer_utils.py
@@ -5,6 +5,7 @@ from eight_mile.pytorch.layers import *
 from eight_mile.optz import create_lr_scheduler
 import baseline.pytorch.embeddings
 import baseline.embeddings
+import random
 from baseline.progress import create_progress_bar
 from torch.utils.data.dataset import IterableDataset, TensorDataset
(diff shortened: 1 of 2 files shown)


api-examples/transformer_utils.py
```diff
@@ -428,12 +432,29 @@ def __iter__(self):
...
+        for file_idx in read_file_order:
+            file = files[file_idx]
             with open(file) as rf:
-                for line in rf:
-                    response = self.process_line(line)
-                    if response:
+                lines = []
+                for i, line in enumerate(rf):
+                    lines.append(line)
+                    if (i+1) % MultiFileLoader.SHUF_BUF_SZ == 0:
```
wenshuoliu, line 446: it seems to me reading the whole file in and shuffle is good enough (for the purpose of reducing core memory usage) and simpler.

dpressel:
we can add back in buffered shuffle if its not efficient, but was simpler to slurp the whole file

PR #649, opened 2020-06-08
feat: allow for setting the max size of n grams in bleu calculations

blester125:
In some uses for seq2seq such as for predicting slot values in a DST task it is possible that your outputs will always has < 4 tokens. 4 is the default
size of n grams used in Bleu. When all of your outputs are < 4 then the precision for 4 grams is always zero which causes the whole bleu score to be
zero. This changes lets you set the max size of n-grams so we can use the bleu scores that pop out of baseline on these sort of datasets.

Tested on:

 Pytorch
 Tf 2 eager mode
 Tf 1 graph mode

What is SOP for testing distributed training?


dpressel:
distributed training should work on single machine, you should be able to test by changing the `--fit_func`

PR #662, opened 2020-06-12
WIP: add tensorboard logging to pytorch pretrain-tlm

dpressel:
also, this adds a logger kwarg to the
CompositeLRScheduler which allows us to
pass in some object that supports the
add_scalar signature provided by the SummaryWriter

if its not passed in, behavior is as before


dpressel:
need a more general solution.  possibly logging from the optimizer manager makes more sense, at least for LR.  also this doesnt work multiworker RN

PR #684, opened 2020-06-26
Feature/pytorch load tfrecord

wenshuoliu:
This PR adds option to read .tfrecord data and train in pytorch. 

This implementation uses `tfrecord.reader.tfrecord_loader` to read file. In this way, it reads multiple files in exactly the same way as the other `MultiFileLoader`. 

I had another implementation that uses `tfrecord.reader.multi_tfrecord_loader`, where reading multiple files relies on `tfrecord` module's implementation. In this way, it's hard to turn off `shuffle`. But IMO `shuffle == False` is only useful when testing response selection performance. I can do it this way if you want.

The real problem here is that I don't observe any speeding up in training.

### DIFF
--- a/api-examples/pretrain_tlm_pytorch.py
+++ b/api-examples/pretrain_tlm_pytorch.py
@@ -103,7 +103,12 @@ def train():
         args.device, updated_local_rank = init_distributed(args.local_rank)
         args.local_rank = updated_local_rank
 
-    reader_type = "lang" if not args.preprocessed else "preprocessed"
+    if args.pattern == '*.tfrecord':
+        reader_type = 'tfrecord'
+    elif args.preprocessed:
+        reader_type = 'preprocessed'
+    else:
+        reader_type = 'lang'
     reader = MultiFileDatasetReader(args.nctx, args.subword_model_file, args.subword_vocab_file, args.pattern,
                                     reader_type=reader_type)
 
--- a/api-examples/pretrain_tlm_pytorch.py
+++ b/api-examples/pretrain_tlm_pytorch.py
@@ -50,7 +50,7 @@ def train():
     parser.add_argument("--num_layers", type=int, default=8, help="Number of layers")
     parser.add_argument("--num_train_workers", type=int, default=4, help="Number train workers")
     parser.add_argument("--nctx", type=int, default=256, help="Max input length")
-    parser.add_argument("--pattern", default='*.json', help="Glob pattern for data")
+    parser.add_argument("--file_type", default='json', help="Glob pattern for data")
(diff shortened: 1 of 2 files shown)


api-examples/transformer_utils.py
```diff
@@ -13,7 +13,7 @@
 from collections import Counter
 import glob
 import json
-
+import tfrecord
```
dpressel, line 16: this creates a dependency in the entire file, do we really want that?

api-examples/pretrain_tlm_pytorch.py
```diff
@@ -102,7 +102,12 @@ def train():
         args.device, updated_local_rank = init_distributed(args.local_rank)
         args.local_rank = updated_local_rank
 
-    reader_type = "lang" if not args.preprocessed else "preprocessed"
+    if args.pattern == '*.tfrecord':
```
dpressel, line 105: In the TF version, I used a file_type, I think that makes more sense than the way we are doing it here now

dpressel: to get the pattern from that is easy
```
    pattern = os.path.join(directory, f'*.{file_type}')
```

api-examples/transformer_utils.py
```diff
@@ -495,6 +498,40 @@ def process_line(self, line):
...
+            for file_idx in read_file_order:
+                file = files[file_idx]
+                idx_file = '.'.join(file.split('.')[:-1]) + '.index'
+                # shard = (worker_id, num_workers), but we already assigned this file to one certain worker,
+                # so shard = (0, 1)
+                itr = tfrecord.reader.tfrecord_loader(file, idx_file, shard=(0, 1))
+                if self.shuffle:
+                    np.random.seed(node_worker_id)
+                    # not sure about the optimal choice of shuffle_queue_size here:
+                    itr = tfrecord.iterator_utils.shuffle_iterator(itr, queue_size=128)
```
dpressel, line 527: seems a bit too small

api-examples/transformer_utils.py
```diff
@@ -517,6 +554,9 @@ def load(self, directory, vocabs, distribute=True, shuffle=True):
         elif reader_type == "lang":
             print("Using files as an LM")
             return SequencePredictionFileLoader(directory, self.pattern, vocabs, self.vectorizer, self.nctx, distribute=distribute, shuffle=shuffle)
+        elif reader_type == 'tfrecord':
+            print("Reading data in .tfrecord fomat using the tfrecord module")
```
dpressel, line 558: spelling

PR #687, opened 2020-06-30
fix: make sure the result from the conv1d is the same shape coming out

blester125:
This PR makes the output size of conv1d's that have even filter sizes the same shape as the input. It does this by padding one less on the front end to match tensorflow.

Tested by training a tensorflow model, converting it into a pytorch model by extracting weights from the checkpoint and comparing the results (both full prediction results as well as internal activation values) between the models. They match. Also tested in cases where the input time dimension was even and when it was odd


dpressel:
It makes more sense to me to change the behavior of `Conv1DSame` and then just use that primitive instead on even filters

PR #701, opened 2020-07-22
fix: bpe/wp taggerservice

blester125:
This PR updates the way we get the labels from the output of the tagger,
we first devectorize the input and check for the token that signals the
subword start if we are in a subword we skip that label.


baseline/services.py
```diff
@@ -481,25 +483,40 @@ def predict(self, tokens, **kwargs):
...
-    def format_output(self, predicted, tokens_batch=None, label_field='label', **kwargs):
+    def format_output(self, predicted, tokens_batch=None, label_field='label', vectorized_examples=None, **kwargs):
+        """This code got very messy dealing with BPE/WP outputs."""
         assert tokens_batch is not None
+        assert vectorized_examples is not None
         outputs = []
+        # Pick a random non-lengths key from the vectorized input, if one input was BPE'd they all had to be to stay aligned
+        key = [k for k in vectorized_examples if not k.endswith("_lengths")][0]
+        # Get the subword start signal from the vectorizer's tokenizer. If there is none we return None
+        subword_start = getattr(getattr(self.vectorizers[key], "tokenizer", None), "subword_start", None)
```
dpressel, line 496: i think this property would be better placed on the actual vectorizer (even if it just delegates to a sub-prop).  this is  a bit hectic here

baseline/services.py
```diff
@@ -481,25 +483,40 @@ def predict(self, tokens, **kwargs):
...
+            vectorized_example = vectorized_examples[key][i]
+            # Convert back into strings, these will now be broken into subwords
+            tokenized_text = [self.rev_vocab[key][t] for t in vectorized_example]
+            new_outcome = []
+            # Loop through all the tokens and collect the labels for the start of subwords
+            for j, t in enumerate(tokenized_text):
+                # If the subword start is None we are a normal vectorizer and want to collect everything
+                # If we start with the subword start we are part of a broken token and our label doesn't matter
+                # If we are the special tags from pre-training our labels don't matter
+                if subword_start is None or (not t.startswith(subword_start) and t not in ("[CLS]", "[SEP]", "[PAD]")):
```
dpressel, line 510: i think the SoC here is somewhat lacking.  wouldnt it be better to ask the vectorizer what its special tokens are?  RN, these are defined as string literals in the middle of the code and these types of items also happen in the vectorizers, why not place them all in the same area?

dpressel: BPE sub-words have @@ at the end: 'Ro@@ asted barr@@ am@@ un@@ di fish', 'Centr@@ ally managed over a cli@@ ent-@@ server architecture']

dpressel: possibly another property in the vectorizer could abstract (`is_leader_token`?)

blester125: Yeah I just noticed that the bpe has a trailing signal instead of a starting one.

Having that be different makes things more complex.

What if the vectorizers had a method that took in a tokenized string (or they could rerun the tokenization but that would be wasteful) and output the indices that were valid tokens?

dpressel: yeah, that would work I guess

baseline/services.py
blester125: This property is now accessible on the vectorizer

PR #709, opened 2020-07-31
make WeightTiedDense lazy via a weight property

dpressel:
this change makes weight tying module use the
tied module's underlying weight directly instead of
copying a pointer parameter.  Previously we were using a
pointer, which causes problems when the module is
rehydrated from disk after first initialization.

This was found because the number of parameters in the
model was increasing after embeddings were reloaded,
which were tied to an LM head.  One other nice property of
deferring to the "held" module is that we dont need properties
instructing us reload the LM head, because this will happen
automatically for models with a tied head, simplifying TLM
(and TiedSeq2Seq) reloads from checkpoints

The only concern with this approach is that there is now
a function powering the `weight` which may cause a bit of latency,
though compared to the matrix multiply this is probably negligable.
Prior to this change however, there was (at a minimum) some memory
that shouldve been freeable that was still being tracked in PyTorch,
and at worst, may have been affecting the NN fine-tuning

### DIFF
--- a/api-examples/generate_mlm.py
+++ b/api-examples/generate_mlm.py
@@ -74,7 +74,7 @@ def create_model(embeddings, d_model, d_ff, num_heads, num_layers, rpr_k, d_k, c
                                                   activation=activation,
                                                   src_keys=['x'], tgt_key='x')
     if checkpoint_name.endswith('npz'):
-        load_tlm_npz(model, checkpoint_name, lm_head=True)
+        load_tlm_npz(model, checkpoint_name)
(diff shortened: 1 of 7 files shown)


blester125:
LGTM

Does something similar need to happen for TF?

dpressel:
> LGTM
> 
> Does something similar need to happen for TF?

I think so, probably need to do the same for `.W`.  Right now it looks like it would have the same issue

dpressel:
The TF seems to work (tested on 2.1) with W as a property.

PR #714, opened 2020-08-08
Feature/label smoothing

blester125:
This PR adds label smoothing for pytorch seq2seq and tf2 eager mode seq2seq. AFAIK we are planning to deprecate tf1 soonish so I didn't bother adding it there.

One weird thing is that we have this `SequenceLoss` wrapper that instantiates a loss function inside of it. We can pass in different losses but if it needs a extra argument it gets weird. I created and inner function that acts like a partial application of the argument.

I also had to add `**kwargs` to all the eager loss function so that I could pass in the label smoothing for the seq2seq one. I tested all the tasks with tf eager mode and they all work

### DIFF
--- a/baseline/tf/seq2seq/training/eager.py
+++ b/baseline/tf/seq2seq/training/eager.py
@@ -13,32 +13,42 @@ from baseline.train import register_training_func, Trainer
 from baseline.tf.seq2seq.training.utils import to_tensors, SHUF_BUF_SZ, NUM_PREFETCH
 
 
-def loss(model, features, labels, **kwargs):
-    # Claims its T, B, H
-    logits = tf.transpose(model(features), [1, 0, 2])
-    # So ok, then transpose this too
-    labels = tf.transpose(labels, [1, 0])
-    # TxB loss mask
-    label_lengths = features['tgt_len']
-    mx_seq_len = tf.reduce_max(label_lengths)-1
-    labels = labels[1:mx_seq_len + 1, :]
-    logits = logits[:mx_seq_len, :, :]
-    label_smoothing = kwargs.get("label_smoothing")
-    if label_smoothing is not None:
+class Seq2SeqLoss(tf.keras.layers.Layer):
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.label_smoothing = kwargs.get("label_smoothing")
+        if self.label_smoothing is not None:
+            self.loss_fn = self.smoothed_loss
+        else:
+            self.loss_fn = self.loss
+
+    def smoothed_loss(self, logits, labels):
         V = get_shape_as_list(logits)[-1]
         one_hot = tf.one_hot(labels, V)
-        losses = tf.keras.losses.categorical_crossentropy(
-            y_true=one_hot, y_pred=logits, from_logits=True, label_smoothing=label_smoothing
+        return tf.keras.losses.categorical_crossentropy(
+            y_true=one_hot, y_pred=logits, from_logits=True, label_smoothing=self.label_smoothing
         )
-    else:
-        losses = tf.nn.sparse_softmax_cross_entropy_with_logits(logits=logits, labels=labels)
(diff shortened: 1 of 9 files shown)


layers/eight_mile/pytorch/layers.py
```diff
@@ -170,6 +170,31 @@ def extra_repr(self):
         return f"reduction={self.avg}"
 
 
+class LabelSmoothingLoss(nn.Module):
```
dpressel, line 173: This should contain documentation and a reference

layers/eight_mile/tf/optz.py
```diff
@@ -776,17 +776,17 @@ def weight_decay_fn():
     def global_step(self):
         return self.optimizer.iterations
 
-    def update(self, model, x, y, num_replicas=1):
+    def update(self, model, x, y, num_replicas=1, **kwargs):
```
dpressel, line 779: there doesnt seem to be much API parity between TF and PyTorch with this feature.  In once case we are changing the impl of EagerOptimizer and in the other its unmodified?

baseline/tf/seq2seq/training/eager.py
```diff
@@ -13,7 +13,7 @@
 from baseline.tf.seq2seq.training.utils import to_tensors, SHUF_BUF_SZ, NUM_PREFETCH
 
 
-def loss(model, features, labels):
+def loss(model, features, labels, **kwargs):
```
dpressel, line 16: I think it would be possible to produce a functor with the right underlying implementation so that this can be done prior to loss function definition (similar to the fact we do this upfront in the PyTorch impl).  Then the loss function could be defined differently depending on if its on or not.  I guess this pushes the problem up but to me it feels weird that the optimizer ends up needing kwargs everywhere to set it up right.  Would be better IMO if we could adding the kwargs to the optimizer and loss function defs

layers/eight_mile/tf/optz.py
blester125: There already wasn't API parity between the two. The pytorch seq2seq model is structured like all out original models (tf1 and pytorch) where there is a `.create_loss` method. TF2 Eager (the only seq2seq supported mode) is already setup differently where there is a `loss` function in the trainer that is responsible for running the model's forward pass and calculating the loss. That function is called by the eager optimizer so the only way to inject things that tweak that loss calculation is to pass them through the EagerOptimzer like this. The TF2 Eager version doesn't have something akin to `.create_loss` that we can modify or override in a subclass.

baseline/tf/seq2seq/training/eager.py
blester125: I link the spirit of this in that it seems to be separating the loss from the model, moving towards something that we are able to specify in the config, swap out, etc but I guess I'm not in love with the loss function being responsible for running the forward part of the model, I know it is mostly there because of how different that tagger was, but I was hoping to not to a huge refactor on how eager losses are done because I still don't know if the way we are doing it is the best way

blester125: Updated to use a functor

baseline/tf/seq2seq/training/eager.py
```diff
@@ -13,24 +13,42 @@
...
-    labels = labels[1:mx_seq_len + 1, :]
-    logits = logits[:mx_seq_len, :, :]
-    losses = tf.nn.sparse_softmax_cross_entropy_with_logits(logits=logits, labels=labels)
-    loss_mask = tf.cast(tf.sequence_mask(label_lengths-1), dtype=tf.float32)
-    losses = losses * tf.transpose(loss_mask, [1, 0])
-
-    losses = tf.reduce_sum(losses)
-    losses /= tf.cast(tf.reduce_sum(label_lengths), tf.float32)
-    return losses
+class Seq2SeqLoss(tf.keras.layers.Layer):
```
dpressel, line 16: It looks like right now the distributed mode hasnt been updated to use this loss but it should be

dpressel: We can add this later. I’ll merge for now to keep it moving

PR #722, opened 2020-08-15
make wordpiece1d more like bpe1d

dpressel:
- adds a flexible begin and end token
- implement count function


dpressel:
> This should be changed according to the `emit_end_toks`, right?

yes, fixed

PR #724, opened 2020-08-16
chore: return scores from the model part of the seq2seq

blester125:
There have been lots of asks about various scoring methods from different mead models. This PR updates the core seq2seq model to return both the predicted output sequences and the associated scores. The main service interface is unchanged, it is just the model itself that is changed.

This changes lets users get scores by subclassing the service if they choose, in the current setup they would need to subclass and retrain a model

Tested by training and running models with `ed_text.py` for tf1, tf2 eager, and pytorch

### DIFF
--- a/baseline/services.py
+++ b/baseline/services.py
@@ -799,7 +799,7 @@ class EncoderDecoderService(Service):
         examples = self.vectorize(tokens_batch)
 
         kwargs['beam'] = int(kwargs.get('beam', K))
-        outcomes = self.model.predict(examples, **kwargs)
+        outcomes, scores = self.model.predict(examples, **kwargs)
         return self.format_output(outcomes, K=K)
 
     def format_output(self, predicted, K=1, **kwargs):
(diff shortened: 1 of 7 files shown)


baseline/services.py
```diff
@@ -799,7 +799,7 @@ def predict(self, tokens, K=1, **kwargs):
         examples = self.vectorize(tokens_batch)
 
         kwargs['beam'] = int(kwargs.get('beam', K))
-        outcomes = self.model.predict(examples, **kwargs)
+        outcomes, scores = self.model.predict(examples, **kwargs)
```
dpressel, line 802: I normally hate doing this but does it make sense to do like Keras and have a flag to decide what to return here?

blester125: I guess that is an option, we are a bit inconsistent but the classifier has always returned classes and scores (on the other had the tagger has never returned scores)

dpressel:
Will merge after fix to `ed_text.py`

PR #728, opened 2020-08-19
chore: added posterior decoding and updating the tagger decoder API and tests

blester125:
While this looks like a big change to the tagger decoders all actually happens in 8mi. This is expending the kind of scores and decoding we can do with the 8mi layers without surfacing them all in mead. This means that if people want to use them they can access them but it doesn't affect the day to day usage of mead.

(It also fixed the warning where a tf tagger can output a B or an I as the last token in a sequence)

I tested exporting ONNX models and the changes to the taggers don't effect the export.

### DIFF
--- a/addons/tagger_score_services.py
+++ b/addons/tagger_score_services.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+
+
+import torch.nn.functional as F
+from baseline.services import TaggerService
+from eight_mile.pytorch.layers import unsort_batch, sequence_mask, MASK_FALSE
+
+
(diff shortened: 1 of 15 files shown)


dpressel:
I am closing in favor of a new branch:

https://github.com/dpressel/mead-baseline/tree/feature/posterior-decoding-no-labels

This version contains several modifications
- It has been rebased against master
- It includes a sample program `tag_text_scores.py` which demonstrates usage
- It fixes a bug in the `dist` path where the input lengths werent being unsorted causing improper scores for tokens

PR #769, opened 2021-01-25
Feature/ft s2s clm

wenshuoliu:
This PR:

- let LM and seq2seq models load pretrained checkpoint through `--checkpoint` of `mead-train`
- move TiedEmbeddingsSeq2SeqModel to core; make the api-example training more like mead training
- remove GOVectorizer and handle prefix/suffix through `emit_begin_tok`, `emit_end_tok`. 
- add acc to seq2seq model testing metrics
- fix the bug that the `tgt_len` was not permuted with the `src_len` (`tgt_len` was used to calculate the bleu score)
- fix `steps_per_epoch` in `pretrain_paired_pytorch.py` due to torch upgrading from 1.4 to 1.7


api-examples/pretrain_paired_pytorch.py
```diff
@@ -178,15 +183,16 @@ def train():
                          logger=logger, checkpoint_name=args.checkpoint)
 
     model.to(args.device)
-    loss_function = model.create_loss(args.loss)
+    if args.model_type=='dual-encoder':
```
dpressel, line 186: this casing isnt required, `create_loss()` has kwargs

wenshuoliu: removed

api-examples/pretrain_paired_pytorch.py
```diff
@@ -80,7 +80,8 @@ def train():
     parser.add_argument("--num_layers", type=int, default=8, help="Number of layers")
     parser.add_argument("--windowed_ra", type=str2bool, default=False, help="whether prevent attention beyond rpr_k")
     parser.add_argument("--num_train_workers", type=int, default=4, help="Number train workers")
-    parser.add_argument("--nctx", type=int, default=256, help="Max input length")
+    parser.add_argument("--src_nctx", type=int, default=256, help="Max input length")
```
dpressel, line 83: I would suggest leaving nctx and offering an optional tgt_nctx.  If its not provided use nctx

api-examples/transformer_seq2seq_response.py
```diff
@@ -6,7 +6,8 @@
 import glob
 from argparse import ArgumentParser
 import baseline
-from transformer_utils import TiedEmbeddingsSeq2SeqModel, find_latest_checkpoint
+from transformer_utils import find_latest_checkpoint
```
dpressel, line 9: I thought we were getting rid of one of these

baseline/pytorch/seq2seq/train.py
```diff
@@ -106,6 +118,7 @@ def _evaluate(self, es, reporting_fns, **kwargs):
             preds.extend(convert_seq2seq_preds(pred, self.tgt_rlut))
             golds.extend(convert_seq2seq_golds(tgt, tgt_lens, self.tgt_rlut))
         metrics = {'bleu': bleu(preds, golds, self.bleu_n_grams)[0]}
+        metrics['acc'] = self._acc(preds, golds)
```
dpressel, line 121: if tgt_lens is one shorter in train and eval, the value passed into convert_seq2seq_golds is probably off by one right?

baseline/reader.py
```diff
@@ -155,7 +155,11 @@ def __init__(self, vectorizers, trim=False, truncate=False):
         self.tgt_vectorizer = None
         for k, vectorizer in vectorizers.items():
             if k == 'tgt':
-                self.tgt_vectorizer = GOVectorizer(vectorizer)
+                self.tgt_vectorizer = vectorizer
+                if not self.tgt_vectorizer.emit_begin_tok:
+                    self.tgt_vectorizer.emit_begin_tok.append('<GO>')
```
dpressel, line 160: would be better if it was `Offsets.VALUES[Offsets.GO]]`

api-examples/pretrain_paired_pytorch.py
```diff
@@ -251,9 +254,16 @@ def train():
             batch = next(train_itr)
             steps += 1
             x, y = batch
-            inputs = x.to(args.device)
-            labels = y.to(args.device)
-            loss = loss_function(inputs, labels)
+            if args.model_type == 'encoder-decoder':
```
dpressel, line 257: it might be cleaner to bind a function pointer like `run_step` to `run_step_s2s` vs `run_step_dual` at initialization time based on model_type

baseline/pytorch/seq2seq/train.py
wenshuoliu: it turns out to be right: in `evaluate()` both `tgt` and `tgt_len` are from `batch_dict`, so the length matches the tgt; in `test()`, both of them from the `input_` (after `make_input`) so they also match. the only difference between the two cases is an `<GO>` at the beginning, which is taken care of inside `convert_seq2seq_golds()`.

api-examples/pretrain_paired_pytorch.py
wenshuoliu: done

baseline/reader.py
wenshuoliu: done

api-examples/pretrain_paired_pytorch.py
wenshuoliu: done

api-examples/transformer_seq2seq_response.py
wenshuoliu: done

PR #780, opened 2021-02-17
Feature/sentence encoders

dpressel:
add sentence encoders (a la SBERT)
    
    add support for fine-tuning transformers on NLI
    data using a dual encoder method where the transformer
    is shared between the NLI sentences and the results
    are concatenated (along with a subtracted vector) and
    sent through a final layer to predict the output classes
    
    for NLI data, this is a 3-class softmax output, but this
    code also supports other numbers of classes.
    
    the approach here is to leverage the existing infrastructure
    in MEAD and 8-mile, building on the existing paired reader support
    `example_type: dual` to get a pair of inputs, and the existing
    Fine-Tuning Models but subclassing this to create a shared embedding
    for the paired input from a single feature.  That model is
    registered as `fine-tune-dual`, and when executed, the paired input
    values are passed in to the EmbeddingsStack, and the normal reduction
    mechanism is used.  In SentenceBERT the pooling operation is a
    combination of the input encoded vectors and a subtracted absolute
    value vector.  That is achieved here through a new reduction:
    `ConcatSubtractReduction`, which is defaulted in the `fine-tune-dual`.


dpressel:
- [ ] Add TF support
- [ ] Test on large NLI corpus

dpressel:
There are problems with TensorFlow deferred mode with this graph.  I believe this is also an issue with paired classification.  The current work-around is always use `--prefer_eager true` when training on these datasets in TF.   Eager mode will become the default soon and eventually deferred support will go a way

PR #810, opened 2021-04-29
use TF.IO for JSON

dpressel:


dpressel:
this gets rid of python function that doesnt run on TPU

PR #822, opened 2021-05-18
WIP: Feature/save vec json

dpressel:
The current vectorizers write out pickle files which make them incompatible with other programming languages.
Most of the behavior of the vectorizers is straightforward to implement in any language, and `mead-export`
supports conversion to popular runtime or serving formats such as TF-serving and ONNX, which leaves the
vectorizers client side.  Due to the approach we have now, the vectorizers would have to be loaded in Python
and possibly converted to some external format.

The change here is to support a JSON-backend format for vectorizers that can be used to allow vectorization
in other programming languages with more minimal hassle.  Additionally the choice of serialization format can
be controlled via the `vec_fmt` property in the mead config (or at the command line with `--x:vec_fmt json|pickle`).

To facilitate these changes, the old save method was factored out into a pickle-specific method, and for these new
files, all of the serialization is done in the vectorizer JSON file.  Pickle supports serialization of objects to be specified
via `__getstate__` and rehydrated with `__setstate__`.   It seemed desirable to support this via the JSON serialization as
well.

We need to test a lot of combinations since any vectorizer could have unique complications using this methodology

### DIFF
--- a/baseline/utils.py
+++ b/baseline/utils.py
@@ -269,7 +269,7 @@ def save_vectorizers_pickle(basedir, vectorizers, name='vectorizers'):
     write_json(vectorizer_modules, module_file)
 
 @export
-def save_vectorizers(basedir, vectorizers, fmt="json", name='vectorizers'):
+def save_vectorizers(basedir, vectorizers, fmt="pickle", name='vectorizers'):
     if fmt == "pickle":
         return save_vectorizers_pickle(basedir, vectorizers, name)
     vec_params = {}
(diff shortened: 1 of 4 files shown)


dpressel:
- [ ] default tagger saves/loads pickle
- [ ] default tagger saves/loads json
- [ ] wordpiece (BERT) classifier saves/loads pickle
- [ ] wordpiece (BERT) classifier saves/loads json
- [ ] BPE TPT tagger saves/loads pickle
- [ ] BPE TPT tagger saves/loads json
- [ ]

dpressel:
Im going to close this for now, as I dont see an immediate pressing need

PR #841, opened 2021-07-23
control when to run validation with do_early_stopping

wenshuoliu:
This fix restores the behavior back to previous when `do_early_stopping=False`.


dpressel:
I think the inner train functionality (starting from the batch object) can be encapsulated into a function, and once that happens, this would clean up the code here and reduce it significantly.  WDYT?

PR #862, opened 2021-08-30
Feature/docker builds

dpressel:


dpressel:
Previously, these builds were done by connecting the build dockerfiles in https://github.com/mead-ml/mead-gpu to the DockerHub builder, but DockerHub has removed this feature for non-paying users.  There was never any great reason to do it the DockerHub way other than simplicity and convenience of setup -- in fact the system was a bit more opaque than using CI/CD from GH

PR #928, opened 2022-06-25
support distillation with soft targets

dpressel:
quickly mock up 1. supporting more losses
in the classifier via kwargs and 2. what
happens if the Y shape is dense.  In that
case the soft target argmax is treated as
the selected label

### DIFF
--- a/baseline/pytorch/classify/train.py
+++ b/baseline/pytorch/classify/train.py
@@ -20,7 +20,12 @@ def _add_to_cm(cm, y, pred):
     if cm is None:
         return
     _, best = pred.max(1)
-    yt = y.cpu().int()
+
+    # If the truth is actually a prob dist, do an argmax RQ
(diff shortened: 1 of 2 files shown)


baseline/pytorch/classify/train.py
```diff
@@ -46,14 +51,14 @@ def __init__(self, model, **kwargs):
         self.optimizer = OptimizerManager(model, **kwargs)
         self.model = model
         if self.gpus > 0 and self.model.gpu:
-            self.crit = model.create_loss().cuda()
+            self.crit = model.create_loss(**kwargs).cuda()
```
dpressel, line 54: need to check if the parameters exist for this function before we jam kwargs in there -- if its a legacy addon, it might not have these args and then it will fail to run.

dpressel: check signature before passing args.  this could get us in trouble if somebody forgets it, but it will at least allow legacy addons to work as before

